PR #12, opened 2020-09-17
Sharing API basic operations implementation

mateusz-garlacz:


cs3api4lab/swagger/share-api.yaml
```diff
@@ -0,0 +1,183 @@
...
+      description: ""
+      produces:
+      - "application/json"
+      parameters:
+      - in: "query"
+        name: "endpoint"
+        description: "endpoint path, default /"
+        required: true
+      - in: "query"
+        name: "file_id"
```
diocas, line 28: Maybe a bit confusing that it's `file_id` and then you give a path.

cs3api4lab/swagger/share-api.yaml
```diff
@@ -0,0 +1,183 @@
...
+    get:
+      tags:
+      - "shares"
+      summary: "Get all given shares"
+      description: ""
+      produces:
+      - "application/json"
+      parameters:
+      - in: "query"
+        name: "user_id"
```
diocas, line 132: Why do you need username? You can only list your own shares.

cs3api4lab/swagger/share-api.yaml
```diff
@@ -0,0 +1,183 @@
...
+    get:
+      tags:
+      - "shares"
+      summary: "Get all received shares"
+      description: ""
+      produces:
+      - "application/json"
+      parameters:
+      - in: "query"
+        name: "user_id"
```
diocas, line 174: Same question as above.

cs3api4lab/config/config_manager.py
```diff
@@ -0,0 +1,42 @@
...
+        config_file = cm.get(self.__config_file_name)
+        self.config = config_file.get("cs3")
+
+        if self.config is None:
+            self.log.error(f'Error while reading cs3 config file {self.__config_dir}{self.__config_file_name}')
+            raise IOError(u'Error while reading cs3 config file')
+        #
+        # Overwriting configuration values with environment variables
+        #
+        env_names = {"reva_host", "client_id", "client_secret", "home_dir"}
```
diocas, line 36: nice!

cs3api4lab/handlers.py
```diff
@@ -2,15 +2,114 @@
...
+    @gen.coroutine
+    def get(self):
+        file_id = self.get_query_argument('file_id', default=None)
+        RequestHandler.handle_request(self, self.share_api.list_grantees_for_file, 200, file_id)
+
+
+handlers = [
+    (r"/api/cs3/shares", ShareHandler),
+    (r"/api/cs3/shares/list", ListSharesHandler),
+    (r"/api/cs3/shares/received", ListReceivedSharesHandler),
```
diocas, line 87: According to your documentation, this should be `list-received`. Please change one or the other.

cs3api4lab/api/cs3_share_api.py
```diff
@@ -0,0 +1,279 @@
...
+        list_request = sharing.ListSharesRequest()
+        list_response = self.gateway_stub.ListShares(request=list_request,
+                                                     metadata=[('x-access-token', self.get_token())])
+        self._check_response_code(list_response)
+        self.log.info("List shares response for user: " + self.config['client_id'])
+        self.log.info(list_request)
+        return self._map_given_shares(list_response)
+
+    def list_grantees_for_file(self, file_id):
+        shares_response = self.list()
```
diocas, line 68: You are listing the shares, you can filter them. Then you don't need to do it yourself. Check `ListSharesRequest.Filter.Type` in the documentation.

cs3api4lab/api/cs3_share_api.py
```diff
@@ -0,0 +1,279 @@
...
+                shares.append(share)
+
+        shares_dict = {}
+        for share in shares:
+            opaque_id = share['grantee']['opaque_id']
+            shares_dict[opaque_id] = share['permissions']
+
+        return shares_dict
+
+    def _map_opaque_id(self, opaque_id):
```
diocas, line 82: What is this?

cs3api4lab/api/cs3_share_api.py
```diff
@@ -0,0 +1,279 @@
...
+        return "user" + str(random.randint(0, 10))
+
+    def _decode_file_id(self, file_id):
+        if '%2F' in file_id:
+            return re.search('(?=%2F).*', file_id).group(0).replace('%2F', '/')
```
diocas, line 87: Shouldn't we do a decode instead?

cs3api4lab/api/cs3_share_api.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def list(self):
+        # todo filters
+        list_request = sharing.ListSharesRequest()
+        list_response = self.gateway_stub.ListShares(request=list_request,
+                                                     metadata=[('x-access-token', self.get_token())])
+        self._check_response_code(list_response)
+        self.log.info("List shares response for user: " + self.config['client_id'])
+        self.log.info(list_request)
+        return self._map_given_shares(list_response)
```
diocas, line 65: I would return a list of shares that map a fileid/etc -> list of people it is shared with.
It makes more sense to have a list of all the unique shares available, instead of returning the repeated values in case it's shared with multiple users.
So I recommend we group by ID/path/some identifier.

diocas (commented):
Just a few comments. Thanks for the PR. 
But I got a bit confused because this has other things not related to the sharing API (probably because the other PR is not approved).

mateusz-garlacz:
> Just a few comments. Thanks for the PR.
> But I got a bit confused because this has other things not related to the sharing API (probably because the other PR is not approved).
Hi, thanks for comments. The changes will be made soon. Some of the changes included in this PR are in fact the result of the other unmerged PR. However this PR also cover commits regarding file API. While working on the sharing API it was necessary to change the project structure and extract common logic to different classes and it was difficult to seperate from specific sharing API tasks. However the logic regarding files did not change, what changed are logic location and variables names.

cs3api4lab/api/cs3_share_api.py
mateusz-garlacz: Sure, it's over-engineered.

diocas:
Tomorrow we should have the other PR merged and I'll do a quick look again and merge this.
Good thing that the filters worked! :) I just betted that it had a better way of retrieving shares and went looking in the API spec, but wasn't sure it would work.

JarCz (approved):

cs3api4lab/api/cs3_share_api.py
```diff
@@ -0,0 +1,364 @@
...
+        list_request = sharing.ListSharesRequest()
+        list_response = self.gateway_stub.ListShares(request=list_request,
+                                                     metadata=[('x-access-token', self.get_token())])
+        self._check_response_code(list_response)
+        self.log.info("List shares response for user: " + self.config['client_id'])
+        self.log.info(list_request)
+        return list_response
+
+    def list_grantees_for_file(self, file_path):
+        stat = self.file_api.stat(file_path, self.config['client_id'])
```
diocas, line 81: There's no way to filter by path? Do we really need to have the resource id?
If that is the case, then the API should expose resource IDs and not paths, as this seems an unnecessary round trip for information that we should already have.
Ofc that dealing with paths makes more sense, though...

cs3api4lab/api/cs3_share_api.py
```diff
@@ -0,0 +1,364 @@
...
+    def list_dir_model(self):
+        list_response = self._list()
+        return self._map_shares_to_dir_model(list_response)
+
+    def _list(self):
+        # todo filters
+        list_request = sharing.ListSharesRequest()
+        list_response = self.gateway_stub.ListShares(request=list_request,
+                                                     metadata=[('x-access-token', self.get_token())])
+        self._check_response_code(list_response)
```
diocas, line 75: If the code is not OK, the only thing we will see in the logs is "code not ok". But then you don't know what you were trying to get in the first place (although you can always check what endpoint was called, but in some other/different log). My point is, you should log what the user is trying to do and wether it failed or not.

diocas: This comment applies to all the other methods.

diocas: Besides, read my comment in `RequestHandler`

cs3api4lab/handlers.py
```diff
@@ -2,15 +2,114 @@
...
+    (r"/api/cs3/shares/file", ListSharesForFile),
+]
+
+
+class RequestHandler(APIHandler):
+
+    @staticmethod
+    def handle_request(self, api_function, success_code, *args):
```
diocas, line 95: Your error handling only produces 200 OKs + requested outputs OR 500. Isn't there a middle term?
Like, file doesn't exist or no permission to change the shares or smth?

diocas: For now let's keep as is and iterate later. But please add a #TODO in the code or an issue in github to followup.

cs3api4lab/api/cs3_share_api.py
```diff
@@ -0,0 +1,364 @@
...
+    auth = None
+    config = {}
+    file_api = None
+    date_fmt = '%d-%b-%Y (%H:%M:%S.%f)'
+
+    def __init__(self, log):
+        self.config = Cs3ConfigManager().config
+        self.auth = Authenticator()
+        self.file_api = Cs3FileApi(log)
+        self.gateway_stub = grpc_gateway.GatewayAPIStub(ChannelConnector().channel)
```
diocas, line 42: why not self.api ?

cs3api4lab/api/cs3apischeckpoint.py
```diff
@@ -1,23 +1,29 @@
 from notebook.services.contents.checkpoints import Checkpoints
```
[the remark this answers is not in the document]
diocas: Why are we changing this?

cs3api4lab/auth/authenticator.py
```diff
@@ -0,0 +1,32 @@
...
+import cs3.gateway.v1beta1.gateway_api_pb2_grpc as cs3gw_grpc
+from cs3api4lab.config.config_manager import Cs3ConfigManager
+from cs3api4lab.auth.channel_connector import ChannelConnector
+
+
+class Authenticator:
+    tokens = {}  # map client_id [string] to {authentication token, token expiration time}
+
+    def __init__(self):
+        self.config = Cs3ConfigManager().config
```
diocas, line 12: Why are you reading the config again? Ans setting another channel? This could be passed by parameter, from the component that has already read it/has it.

I understand we are trying to make these things independent, but I think this shouldn't be.
We should't keep reading the same thing again...

cs3api4lab/auth/channel_connector.py
```diff
@@ -0,0 +1,32 @@
...
+class ChannelConnector(LoggingConfigurable):
+    channel = None
+
+    def __new__(cls, *args, **kwargs):
+        if cls.channel is None:
+            cls.channel = super(ChannelConnector, cls).__new__(cls)
+        return cls.channel
+
+    def __init__(self):
+        config = Cs3ConfigManager().config
```
diocas, line 16: Again with reading the config instead of re-using

diocas (changes requested):
Please add documentation. Is really hard to understand why we need to decode of purify stuff and why things are done the way they are.

cs3api4lab/api/cs3_share_api.py
mateusz-garlacz: Initially all of the shares were fetched and filtered manually by file path. With your advice I used a built-in CS3 Api filter, which made more sense. However to retrieve a share using a filter, the resource ID including storage ID and file ID is needed. Unfortunatelly it turned out, that the frontend does not have the knowledge about storage IDs. The request containes only file path as it was before, but the 'round-trip' is necessary to make it work if we want to use the filter

cs3api4lab/handlers.py
mateusz-garlacz: I added an issue and will take care of that after this PR. I will add additional codes, however it is worth noting that reva itself returns only 200s and 500s on most of the errors

cs3api4lab/api/cs3apischeckpoint.py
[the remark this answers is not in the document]
mateusz-garlacz: It was added long time ago and forgotten. I will delete it as it provides no functionality and as I understand checkpoints are not the topic for now.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+    jupyter lab build -y && \
+    jupyter lab clean -y && \
+	npm cache clean --force && \
+    rm -rf "/home/${NB_USER}/.cache/yarn" && \
+    rm -rf "/home/${NB_USER}/.node-gyp" && \
+    fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/opt/cs3" && \
+    fix-permissions "/home/${NB_USER}" && \
+	sed -i 's/#c.NotebookApp.contents_manager_class/c.NotebookApp.contents_manager_class/g' /home/${NB_USER}/.jupyter/jupyter_notebook_config.py && \
+	sed -i 's/notebook.services.contents.largefilemanager.LargeFileManager/cs3api4lab.api.cs3apismanager.CS3APIsManager/g' /home/${NB_USER}/.jupyter/jupyter_notebook_config.py
```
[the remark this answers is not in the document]
mateusz-garlacz: This PR includes changes in README and mentions our CS3APIsManager implementation as far as I can see?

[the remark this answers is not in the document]
diocas: I'm just asking why you need to point to `cs3api4lab.api.cs3apismanager.CS3APIsManager` when this looks to be included in `__init__` already, so you should be able to just do `cs3api4lab.CS3APIsManager`, right?

cs3api4lab/api/cs3_share_api.py
diocas: Ok, let's keep it like this for now. I would say that the front end needs to know the ID, so this will have to be added. (Or we should ask for the CS3APIs to support filter by path.
Put a `FIXME` in the code so that we show that we know this is not great.

cs3api4lab/handlers.py
diocas: Take note of all these limitations. We should give our findings as a list of improvements to the CS3APIs!

Dockerfile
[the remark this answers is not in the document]
JarCz: We need to replace the default content class in JupyterLab (LargeFileManager) with our content class implementation. If we don't change the config file, JupyterLab load extension but don't replace the default content class.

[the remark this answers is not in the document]
diocas: Guys, I'm just saying that since we do `from cs3api4lab.api.cs3apismanager import CS3APIsManager` inside `cs3api4lab/__init__.py`,  we don't need to do `c.NotebookApp.contents_manager_class = cs3api4lab.api.cs3apismanager.CS3APIsManager` because `c.NotebookApp.contents_manager_class = cs3api4lab.CS3APIsManager` is enough.
Is enough, simpler, hides the internal organization of the code and we don't need to change the README.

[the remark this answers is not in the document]
mateusz-garlacz: Ok, got it

diocas:
@mateusz-garlacz Please let me know when this is ready to be reviewed and merged. Cannot promise I'll do it this week, though.

diocas (approved):

PR #17, opened 2020-10-22
Basic authentication structure and interceptor

### LINKED ISSUE #10: Authentication mechanism
This issue is to discuss the authentication of the extension, as requested in the last meeting.

## 1. Context
Allowing the use of username+password can be maintained in the config, although having passwords in plain text is not recommended. But this might be useful for local dev (we should give a warning if the user uses this).

The best way to authenticate should be via OIDC/oAuth tokens. These tokens expire so we need to allow a configurable way of updating them (since different authentication methods might have different ways of doing things). For SWAN, for example, it might be better that we read the EOS oAuth file*, since it's already updated and is being used by the FUSE mount. For other installations without EOS, maybe calling an external service or asking JupyterHub might be the solution (which is how in SWAN we renew the EOS oAuth file: we retrieve the latest auth information from JH** and write to a file. For SWAN I propose reading this file to avoid having 2 processes asking JH for the same information).

## 2. Implementation
I suggest adding a configuration to our ContentsManager that receives a method. This method should receive the `userid` + the current - in use - `token`, and will return a new `token` that should be used instead. Then we can call this method right before we create the `AuthenticateRequest`.
Why do we pass the current token? Because the renewal might need some extra information from it, or it can check wether it's still valid.

We can provide some example functions, like one that reads from JH**, another that reads a file and another that reads the EOS special format file*. This way, in the configuration we don't need to implement these basic cases, we just import and use one of those.

@JarCz please let me know if you need any clarification!

---
\* The EOS oAuth file has the following format: "_oauth2:<OAUTH_TOKEN>:<OAUTH_INSPECTION_ENDPOINT>_".

** If you're using [oauthenticator](https://github.com/jupyterhub/oauthenticator), you can access the user oauth token by querying the user endpoint of JupyterHub, like so (from inside the Jupyter terminal):
```bash
$ curl -H "Authorization: token $JUPYTERHUB_API_TOKEN" $JUPYTERHUB_API_URL/users/$NB_USER
```
```json
{
    "kind": "user",
    "name": "username",
    "admin": true,
    "groups": [],
    "server": "/user/username/",
    "pending": null,
    "created": "2020-08-31T09:31:24.638244Z",
    "last_activity": "2020-08-31T10:43:46.033379Z",
    "servers": {
        "": {
            (...)
        }
    },
    "auth_state": {
        "access_token": "blablabla",
        "refresh_token": "blablabla",
        "oauth_user": {
            "sub": "username",
            "cern_uid": 00000,
            "preferred_username": "username",
            "cern_gid": 00000,
            "given_name": "Diogo",
            "cern_preferred_language": "EN",
            "home_institute": "CERN",
            "cern_person_id": 821079,
            "name": "Diogo Castro",
            "cern_mail_upn": "mail@mail.com",
            "family_name": "Castro",
            "email": "my.email@mail.com",
            "cern_upn": "username"
        },
        "scope": [
            "oidc-cern-profile",
            "oidc-email",
            "oidc-cern-login-info",
            "oidc-client-id"
        ]
    }
}
```
In our KeyCloak authenticator, based on `oauthenticator`, we implement the JupyterHub function that updates this information automatically. But this is not implemented by default by the other `oauthenticator`'s.

JarCz:
Closes #10


cs3api4lab/auth/check_auth_interceptor.py
```diff
@@ -0,0 +1,39 @@
...
+    def intercept_stream_unary(self, continuation, client_call_details, request_iterator):
+        return self._intercept_call(continuation, client_call_details, request_iterator)
+
+    def intercept_unary_stream(self, continuation, client_call_details, request):
+        return self._intercept_call(continuation, client_call_details, request)
+
+    def intercept_stream_stream(self, continuation, client_call_details, request_iterator):
+        return self._intercept_call(continuation, client_call_details, request_iterator)
+
+    def _intercept_call(self, continuation, client_call_details, request_or_iterator):
```
diocas, line 30: I'm not sure how this works. Please provide info.

But.. I would intercept calls to REVA and:
1. check if the current token is still valid (by checking the expiration), if not refresh it. Only after that I would call the function we're wrapping.
2. if the call to reva failed due to auth problems, we refresh the token AND try again. This would allow us to try one more time and thus avoid having to check this in other parts of the code (or worse, sending the error to the ui).

cs3api4lab/api/cs3_share_api.py
```diff
@@ -37,12 +39,15 @@ class Cs3ShareApi:
...
-        channel = ChannelConnector.get_channel()
-        self.config = config
-        self.auth = Authenticator(config, channel)
+        self.config = Cs3ConfigManager().get_config()
+        self.auth = Authenticator(config=self.config, log=self.log).instance
+
         self.file_api = Cs3FileApi(log)
-        self.cs3_api = grpc_gateway.GatewayAPIStub(channel)
+
+        channel = ChannelConnector().get_channel()
```
diocas, line 47: Why don't we already return the intercepted channel in `ChannelConnector().get_channel()`?

cs3api4lab/auth/eos.py
```diff
@@ -0,0 +1,38 @@
...
+        elif "eos_token" in self.config and self.config['eos_token'] is not "":
+            eos_token = self.config['eos_token']
+        else:
+            raise AttributeError("Config haven't EOS token or token file.")
+
+        self.tokens = eos_token.split(':')
+        if len(self.tokens) != 3 or self.tokens[0] != 'oauth2':
+            raise AttributeError("It's not valid EOS token.")
+
+    def authenticate(self, client_id):
```
diocas, line 37: Should we raise an exception if the token already expired?

cs3api4lab/auth/eos.py
```diff
@@ -0,0 +1,38 @@
...
+class Eos(Authenticator):
+    tokens = {}
+
+    def __init__(self, config=None, log=None):
+
+        self.config = config
+        file = None
+        if "eos_file" in self.config and self.config['eos_file'] is not "":
+
+            try:
```
diocas, line 19: https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files

> It is good practice to use the with keyword when dealing with file objects. The advantage is that the file is properly closed after its suite finishes, even if an exception is raised at some point. Using with is also much shorter than writing equivalent try-finally blocks:

cs3api4lab/auth/eos.py
```diff
@@ -0,0 +1,38 @@
...
+from jupyter_core.paths import jupyter_config_path
+from notebook.services.config import ConfigManager
+
+from cs3api4lab.auth.authenticator import Authenticator
+
+
+class Eos(Authenticator):
+    tokens = {}
+
+    def __init__(self, config=None, log=None):
```
diocas, line 13: Since I said we should properly inherit from the parent, I would call the parent init and let if set the config.

cs3api4lab/auth/eos.py
```diff
@@ -0,0 +1,38 @@
...
+
+        self.config = config
+        file = None
```
diocas, line 16: No need to have file declared, see bellow.

cs3api4lab/auth/eos.py
```diff
@@ -0,0 +1,38 @@
...
+            except IOError as e:
+                raise IOError(f"Error opening token file {self.config['eos_file']} exception: {e}")
+            finally:
+                if file is not None:
+                    file.close()
+
+        elif "eos_token" in self.config and self.config['eos_token'] is not "":
+            eos_token = self.config['eos_token']
+        else:
+            raise AttributeError("Config haven't EOS token or token file.")
```
diocas, line 31: "hasn't"

cs3api4lab/auth/eos.py
diocas: And since when is this authenticating? We still need to login to reva, right? (see my comment in the review)

diocas: Also, oAuth tokens expire. This is the only reason why we have a EOS file in the first place (because we have another process that is refreshing and writing the updated token to that file).
So, if we are authenticating we need to read the file again and use the new token. Please remove that logic from the init and put it here.

cs3api4lab/auth/reva_token.py
```diff
@@ -0,0 +1,18 @@
+from tornado import web
+
+from cs3api4lab.auth.authenticator import Authenticator
+
+
+class RevaTokenAuth(Authenticator):
```
diocas, line 6: I don't see a reason for having this atm.

cs3api4lab/auth/reva_password.py
```diff
@@ -0,0 +1,40 @@
...
+import cs3.gateway.v1beta1.gateway_api_pb2_grpc as cs3gw_grpc
+import cs3.rpc.code_pb2 as cs3code
+import time
+from tornado import web
+
+from cs3api4lab.auth.authenticator import Authenticator
+from cs3api4lab.auth.channel_connector import ChannelConnector
+
+
+class RevaPassword(Authenticator):
```
diocas, line 11: Since the EOS/oAuth token also needs to be converted in a REVA token (by calling the `Authenticate`, as I mention in my review), we should re-use this code. So I would put it in the parent.

diocas (changes requested):
Thanks for the PR. Made a few comments.
It would be much easier to review if you explained in the PR what you are doing and why you are doing it. And having comments in the code would also help. Otherwise, it's taking me a lot of time to understand things and then go into the code itself.

I'm pretty sure I've mentioned it in previous reviews, but I believe this is not how you do oAuth authentication to REVA. You still need to login, as if it was a username/password. But, the type is `bearer`, the clientid is the clientid and the secret is the oauth token (instead of the pwd). Please correct me if I'm wrong, but that's what I recall.

So I would move all the logic of the the cs3 `Authenticate` to the parent class and call it from the children with the correct parameters (including the logic for storing the different user tokens). This also means that you set the type internally, instead of getting this from the config, as you do for the RevaPassword.
The children classes will only be responsible for getting the username/pass combo from the config file and call authenticate with them (in the RevaPassword case), or responsible for reading the token/token file and call authenticate with them. Nothing more.

Also, I'm not seeing the need for having RevaToken so I would remove it. Otherwise what I just said above would need some extra work. Let me know if you disagree.

On the interceptor, I would check if the auth failed, refresh accordingly and re-try, if this is possible.

diocas:
I forgot to mention.. We should have an Authenticator which is `oAuth`. The EOS Authenticator will inherit from it, strip the token from the file and then pass it.
Otherwise we don't have a generic thing and only have the EOS specific.

README.md
```diff
@@ -145,6 +145,60 @@ Config file fields:
...
+  * Reva user and secret:
+ 
+ ```json
+{
+  "cs3":{
+    ...
+	"authenticator_class": "cs3api4lab.auth.reva_password.RevaPassword",
+	"client_id": "einstein",
+	"client_secret": "relativity",
+	"login_type": "basic"
```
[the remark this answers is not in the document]
JarCz: We can remove this value and add the default type, but in Riva config, we can configure different login types.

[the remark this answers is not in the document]
diocas: If you want to make it more generic, make it the default value but with the possibility to be changed.

cs3api4lab/auth/authenticator.py
```diff
@@ -1,31 +1,47 @@
...
+            class_name = self.config['authenticator_class'].split('.')[-1]
+            module_name = self.config['authenticator_class'].split(class_name)[0]
+            module_name = module_name.rstrip('.')
+
+            if class_name == "Authenticator":
+                raise AttributeError("Can't instance Authenticator class")
+
+            module = importlib.import_module(module_name)
+            clazz = getattr(module, class_name)
+            self.instance = clazz(config=self.config, log=log)
```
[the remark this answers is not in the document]
JarCz: At this point, I'm using a different way to create instance a class because It is generating from the config value. I didn't remember, but I tried to using typical instance class from JupyterLab (trait type). We have parent classes with basic methods. Each children classes must implement the unique code.

cs3api4lab/auth/check_auth_interceptor.py
JarCz: The interceptor is a second layer of checking auth token, first is in the specific authentication class. 
For examples: 

- RevaPassword doesn't check auth token because getting expiration time form config file (like a WOPI example) and refreshes token if the expiration time is over
- RevaToken checking expiration time before send a request to the Riva, but can't refresh token and throw Http 401 error if a token is expired
-  EOS type now doesn't check expiration time and refresh token, but can when I will finish the mock server and create code to refresh token (next PR)

In future, we need to add code to resend request if a token is a refresh, but it's not trivial because we will have some probability of conflict situations. (New issue ?)

cs3api4lab/api/cs3_share_api.py
JarCz: Because, channels it is a generic type using for authentication, .etc, intercept channel is specific type using in Riva data operation (stat, read a file, write a file, .etc).

cs3api4lab/auth/eos.py
JarCz: If the token is expired we have a problematic situation: refresh from file or refresh from oAuth service (endpoint in the token). Which method is preferred to use?

cs3api4lab/auth/reva_token.py
JarCz: The WOPI has two types of auth with user/pass and token (UserID in method). Now I can't say if token auth will be using, but another project has this functionality. Example: https://github.com/cs3org/wopiserver/blob/0b6f9198e5d3629e3d5903b305ad9b85d51e628b/src/cs3iface.py#L40

cs3api4lab/auth/reva_password.py
JarCz: Is the EOS token equals to Riva token?
If not this is topic to discuss on the weekly meeting because Riva token is similar oAuth with extra information (below is an example of the decoded token). In my opinion, If we need to change the EOS token to Reva token, I can't do it this from the plugin side, (except if EOS token has a username and secret) only from Riva side. Maybe we need a new auth provider. This auth provider gets the EOS token and returns Riva auth token.

`
{
  "aud": "reva",
  "exp": 1602145654,
  "iat": 1602059254,
  "iss": "cernbox.cern.ch",
  "user": {
    "id": {
      "idp": "cernbox.cern.ch",
      "opaque_id": "4c510ada-c86b-4815-8820-......."
    },
    "username": "einstein",
    "mail": "einstein@......",
    "display_name": "Albert Einstein",
    "groups": [
      "sailing-lovers",
      "violin-haters",
      "physics-lovers"
    ]
  }
}
`

cs3api4lab/api/cs3_share_api.py
diocas: Ok, understood

cs3api4lab/auth/check_auth_interceptor.py
```diff
@@ -0,0 +1,39 @@
...
+                           grpc.UnaryStreamClientInterceptor,
+                           grpc.StreamUnaryClientInterceptor,
+                           grpc.StreamStreamClientInterceptor):
+    unauth_codes = {cs3code.CODE_UNAUTHENTICATED}
+
+    def __init__(self, log, authenticator):
+        self.log = log
+        self.authenticator = authenticator
+
+    def intercept_unary_unary(self, continuation, client_call_details, request):
```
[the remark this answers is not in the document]
diocas: Please resolve the issues if you fixed them. If you disagree with what I say, reply.

cs3api4lab/auth/check_auth_interceptor.py
diocas: I still don't understand why you need a mock server for EOS (unless you're trying to do something else).. This has nothing to do with EOS, is generic oAuth. The only thing specific to EOS is the file contents format which has other elements besides the banal token.

I can create an issue with the retry. But I don't understand why we have 2 layers of checks. Having 2 layers makes the retrial in the interceptor unnecessary because the token won't be refreshed here.

There's something in this logic that I think is not going well... As I suggested from the beginning, we should call the check/refresh token in the authenticator, and that's it. If it fails, we return immediately a 401. If it needs to renew, we renew.
Since these authenticators should inherit from a generic class that does a "login" agains REVA, it means that the check will do the following:

1. Is the reva token ok? if so, return
1. if oauth
    1. is the oauth token valid? If so, call login again and return new token
    1. else raise 401
1. if user/pass
    1.  call login again and return new token

cs3api4lab/api/cs3_share_api.py
diocas: Ok, makes sense.

cs3api4lab/auth/eos.py
diocas: Read from the file the new token. If it's not valid, give 401.

cs3api4lab/auth/eos.py
diocas: Do not resolve issues without a fix.

cs3api4lab/auth/reva_token.py
diocas: I don't understand. The link you sent if for testing and is username/password. But still, for now I don't see a need to have reva tokens because we're not being called by reva.

cs3api4lab/auth/reva_password.py
diocas: Technically there is the functionality to get EOS token, but I never talked about that. For this context, there are no EOS tokens. EOS uses oAuth... The EOS file has extra info, but the token itself is pure oAuth, as retrieved from the idp.

But afaik, you always need to exchange oAuth tokens with the REVA ones. That is why I'm saying you need to authenticate as bearer. If you use the revacmd, you will see that you can login with basic or bearer and, if you choose the bearer, you can give the oAuth token as password.

cs3api4lab/auth/check_auth_interceptor.py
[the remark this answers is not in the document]
JarCz: I've added a comment to the auth.py file

cs3api4lab/auth/check_auth_interceptor.py
JarCz: If eos token is a similar oAuth token I'm using mock to creating /refreshing oAuth token because: 
- I'm don't have a test system with the OAuth server like Keycloak. 
- Configure Keycloakor or another service is something like "cannon on flies". 
- In future, we can use a mock server to mock other users case. 
- Example of "jupyterhub/oauthenticator" mock all request/response to external services.

Interceptor is a good way to check general errors in one place (now for an example 401) whit SOLID, DRY rule. Now isn't general checker because expiry time in the token is checked before request.

This is slowly turning to an architectural problem :). How much business logic parent class must have? For me, it's the parent class is thin it's better. Each specialized code I would like to move to inherit classes.

Ok, I will prepare the auth classes with these rules.

cs3api4lab/auth/eos.py
JarCz: It is fixed in the last commit. Line 25 and 26 in cs3api4lab/auth/eos.py

cs3api4lab/auth/check_auth_interceptor.py
diocas: (Just saying again, there is not EOS token in this context, just oAuth written to a file.)

I like interceptors, I just don't see any utility at the moment in this implementation. Checking the validity of the tokens should be done before, in a place where we can act if there was any expiration. 

I understand and agree that the parent class should be thin. But, at the moment the authentication uses always the same logic, which is doing a `AuthenticateRequest` to REVA with 3 parameters. *Always*. If you put this logic in the children, then you have a lot of duplication.
In the future, we might have different methods to authenticate to REVA, but right now we only have one.

cs3api4lab/auth/eos.py
diocas: True. My bad!

cs3api4lab/auth/reva_token.py
JarCz: I removed Riva token auth method.

cs3api4lab/auth/check_auth_interceptor.py
JarCz: In the last commit, I'm refactored and moved the basic auth method to the parent class.

cs3api4lab/auth/reva_password.py
JarCz: I added the oAuth auth class.

cs3api4lab/auth/authenticator.py
```diff
@@ -1,31 +1,100 @@
...
-        # piggy back on the opportunity to expire old tokens, but asynchronously
-        # as to not impact the current session: let's use python3.7's coroutines support
-        # asyncio.run(_async_cleanuptokens())
-
-        return self.tokens[client_id]['tok']
+        """
+        The basic authenticate method, return IOP token or refresh if is not present or invalid.
+        """
+        if client_id not in self.tokens or (client_id in self.tokens and self._check_token(self.tokens[client_id])):
+            self.refresh_token(client_id)
```
diocas, line 38: Why do you keep passing `client_id` all over the place if this is part of `self.config`?

cs3api4lab/auth/reva_password.py
```diff
@@ -0,0 +1,10 @@
+from cs3api4lab.auth.authenticator import Authenticator
+
+
+class RevaPassword(Authenticator):
+
+    def __init__(self, config=None, log=None):
+        super().__init__(config, log)
+
+    def refresh_token(self, client_id=None, client_secret_or_token=None):
```
diocas, line 9: very confusing.. `client_id` as parameter, but then we use the `client_id` from config.. But then this method can be called without anything. I'm not able to follow up.

cs3api4lab/auth/check_auth_interceptor.py
```diff
@@ -0,0 +1,39 @@
...
+    def _intercept_call(self, continuation, client_call_details, request_or_iterator):
+        response = continuation(client_call_details, request_or_iterator)
+        self._check_result(response.code(), response.result())
+        return response
+
+    def _check_result(self, code, result):
+        if result is not None and result.status is not None and result.status.code in self.unauth_codes:
+            if self.log is not None:
+                self.log.info("Refresh auth token")
+            self.authenticator.refresh_token()
```
diocas, line 39: I still don't understand why we intercept things if we don't re-issue the command.  But then you call refresh without the `client_id`.. So, in practice, this doesn't do anything?

cs3api4lab/auth/authenticator.py
JarCz: I don't remember "Why". In the first code version, I'd like to gets client_id from get/post request. I change it.

diocas: I think your initial idea was that this could be made a library in the future and used by other clients that might authenticate with different users. But I'm still not 100% this will be needed... In lab for sure it's not, so I would agree with a clean up on this.

JarCz: 👌

cs3api4lab/auth/check_auth_interceptor.py
JarCz: I raise 401 error when Reva return unauth code instead call method to refreshing token

cs3api4lab/auth/reva_password.py
JarCz: I did remove client_id in public methods for auth, share and file API

cs3api4lab/auth/check_auth_interceptor.py
diocas: I don't understand your message. Can you explain?

JarCz: Now, if we checked tokens expire before making a request and refresh if is necessary. In this case, it very rare to have a response from Reva with an authentication error. If so, it is better to return 401 than try to refresh the token and add request repetition logic.

diocas: But we do check them, no? When you ask for the token, it's validated before being returned, so we don't need this.

```python
    def authenticate(self):
        """
        The basic authenticate method, return IOP token or refresh if is not present or invalid.
        """
        if self.token is None or self._check_token(self.token) is False:
            self.refresh_token()

        return self.token
```

diocas: So yes, return a 401 if it failed and that's it.

JarCz: In general "yes". It already depends on the implementation of "refresh_token" in the children class. For example, RevaPassword use values from config to log in at Reva, in failure raise 401 error; oAuth has a different way: check token, is not valid raise 401 error, login at Reva, in failure raise 401 error. If classes don't override "refresh_token" method immediately raise 401 error.

cs3api4lab/auth/__init__.py
```diff
@@ -0,0 +1,3 @@
+from cs3api4lab.auth.eos import Eos
```
diocas, line 1: Don't you need to import `check_auth_interceptor` here?

In the files and shares api you do `from cs3api4lab.auth import check_auth_interceptor`, which looks like it will fail.

diocas (dismissed):
Just a question. Will merge after.
Thanks

cs3api4lab/auth/__init__.py
JarCz: I add this import and fix broken dependency to "pyjwt".
But, the plugin works without this import when I tested and local instance.

diocas (approved):

PR #18, opened 2020-10-23
added new exception types and diversified error codes

mateusz-garlacz:


diocas:
@mateusz-garlacz there's really no other information from the api? You always check if it's `cs3_code.CODE_OK`, otherwise you give an error. But there are many possible causes for an error.
If that is the case, we should ask for more kinds of errors.

mateusz-garlacz:
> @mateusz-garlacz there's really no other information from the api? You always check if it's `cs3_code.CODE_OK`, otherwise you give an error. But there are many possible causes for an error.
> If that is the case, we should ask for more kinds of errors.

@diocas well in fact there are many possible error codes coming from Reva, however in practice the response is binary and the incoming codes are OK and ERROR_CODE_INTERNAL regardless of the reason. Would you suggest creating an issue regarding this?

diocas:
Yes, definitely. I was looking at the list and some of the codes would already help us distinguish the errors. 
If you tried simulating the errors and still got a generic code, then we should ask for more appropriate ones.
Should I still merge this PR?

mateusz-garlacz:
Yes, please merge it anyway. It's still better than no distinction at all. I will get back to it after Reva team intruduces the changes.

diocas (approved):

PR #24, opened 2020-11-03
Configuration files location

### LINKED ISSUE #22: Change the location of the config files
(this is a summary of my email from 16th Oct; nothing new here)

When this plugin is used in Docker in production, the user home dir (`/home/jovyan`) will be always an externally-mounted Docker volume. The contents we add to /home/jovyan in Dockerfile will not be accessible. Moreover, the contents of this directory is fully managed by a user. This happens for instance when running in Kubernetes, so then we experience that the config files are missing. 

Conclusion: We should keep the config files elsewhere.

**jupyter_notebook_config.py**
As far I understand, in this file, we need it only to set `c.NotebookApp.contents_manager_class`. According to [the docs](https://jupyter.readthedocs.io/en/latest/use/jupyter-directories.html#configuration-files), the config file might be in `~/.jupyter/`, but if it's not there (i.e. when the user does not override the configuration), other locations are checked, such as `/etc/jupyter/jupyter_notebook_config.py`.

And in the base image we use ([jupyter/base-notebook](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile)), the base config is actually in `/etc/jupyter/jupyter_notebook_config.py`. Couldn't we append this file in Dockerfile, instead of creating a new one in `~/.jupyter`?

**jupyter_cs3_config.json**
This file contains REVA settings, and a user has no control over such settings. Thus, this json file should be put to other location that will be not modified by the user. For instance in `/etc/jupyter`, for consistency.

mateusz-garlacz:
Changes: overwriting contents manager in config file already existing in the container, extracted reva configuration out of docker image to read the config from the environment instead - reagarding #22 

closes #22


diocas (approved):
Approving.
I liked the previous way where we could have a default config, but I guess this will lead to less confusion.

Issue #27, opened 2020-11-16
Remove variable CS3_EXT_CONFIG


michzimny:
Configuration from environment variables is only read when `CS3_EXT_CONFIG` is set `true`. 

But configuration from environment variables should be always read from env variables if they are set. If one runs a Docker container with some env vars set, it is expected that the config is read by the application. 

In addition, isn't setting `CS3_EXT_CONFIG=false` (the default) rather a development setting?

It's closely related to #26 but can be also solved independently (if only one of these suggestions gets accepted).

PR #35, opened 2020-11-25
Running docker image with required vars

### LINKED ISSUE #26: Do not require setting all environment variables
When running in Docker, it is required to set all ([currently ca. 13](https://github.com/sciencemesh/cs3api4lab/blob/master/reva_config.env)) environment variables for a container. If any variable is missing, the error is raised:

```
raise ValueError(f"Unable to read {name} from environment variables")
```

In this approach, if a new env variable was added, it would be always a backward-incompatible change as it would require users of the Docker image to adopt their scripts and provide a value for the new variable.

It is widely-adopted practice to have default values for all (most of?) the variables specified in the Docker image. Then, when running a container, one would only set a few variables that need to be changed from the defaults.

mateusz-garlacz:
So now to run the image only reva host and credentials are required. The rest of the values have default value, but are overwritable. @michzimny @diocas 
Will close #26 and #27.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1578,9 +1578,9 @@
       "dev": true
     },
     "node-fetch": {
-      "version": "2.6.0",
-      "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.6.0.tgz",
-      "integrity": "sha512-8dG4H5ujfvFiqDmVu9fQ5bOHUC15JMjMY/Zumv26oOvvVJjM67KF8koCWIabKQ1GJIa9r2mMZscBq/TbdOcmNA=="
+      "version": "2.6.1",
+      "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.6.1.tgz",
+      "integrity": "sha512-V4aYg89jEoVRxRb2fJdAg8FHvI7cEyYdVAh94HH0UIK8oJxUfkjlDQN9RbMx+bEjP7+ggMiFRprSti032Oipxw=="
     },
     "normalize.css": {
       "version": "8.0.1",


Dockerfile
```diff
@@ -38,7 +38,13 @@ RUN cd /opt/cs3 && \
 
 RUN fix-permissions "/home/${NB_USER}"
 
-ENV JUPYTER_ENABLE_LAB = 1
```
diocas, line 41: The idea was not to set the env variables in the image but to start the extension with default configs.
The difference is that the extension should not require these extra configs even if installed in a different docker image (which will be the case)

README.md
```diff
@@ -89,9 +89,7 @@ git clone https://github.com/cs3org/reva
 cd reva
 make deps
 make
-mkdir -p /var/tmp/reva
-cd examples/ocmd/
-../../cmd/revad/revad -c ocmd-server-1.toml
+cd examples/ocmd/ && mkdir -p /tmp/reva && && mkdir -p /var/tmp/reva
```
diocas, line 92: Please rebase. Michal had changed this.

cs3api4lab/config/config_manager.py
```diff
@@ -9,34 +9,23 @@ class Config(LoggingConfigurable):
...
-                   "secure_channel",
-                   "client_cert",
-                   "client_key",
-                   "ca_cert",
-                   "client_id",
-                   "client_secret",
-                   "login_type"}
+    __env_vars_req = {"reva_host",
+                      "client_id",
+                      "client_secret"}
```
diocas, line 14: As I mentioned in #26, the secret won't be necessary in the configuration if we have different authenticators (this is still not merged).
So maybe we should't check it here.

cs3api4lab/config/config_manager.py
```diff
@@ -9,34 +9,23 @@ class Config(LoggingConfigurable):
...
-            self.log.info("Reading CS3 config from environment")
-            for name in self.__env_names:
-                env_name = "CS3_" + name.upper()
-                if env_name in os.environ:
-                    self.log.info(f"Reading config value {name} = " + str(os.environ[env_name]))
-                    self.config[name] = os.environ[env_name]
-                else:
-                    self.log.error(f"Unable to read {name}")
-                    raise ValueError(f"Unable to read {name} from environment variables")
+        if 'CS3_ENV_CONF' in os.environ.keys() and utils.strtobool(os.environ['CS3_ENV_CONF']):
```
diocas, line 26: We shouldn't use `CS3_ENV_CONF`. This is more os less the same as having `CS3_EXT_CONFIG`.
My proposal is to return to the previous behavior: 
1. we start by loading default values, 
1. if the configuration file exists we load it and replace the values,
1. and if ENV vars exists, we also replace the corresponding values.

diocas (approved):
I've added some comments of things to change. @michzimny do you have anything else?

README.md
```diff
@@ -160,10 +158,22 @@ Build docker image:
...
+- CS3_CLIENT_ID - client login to authenticate in Reva [required]
+- CS3_CLIENT_SECRET - client password to authenticate in Reva [required]
+- CS3_AUTH_TOKEN_VALIDITY - the lifetime of the authenticating token
+- CS3_ENDPOINT - endpoint for Reva storage provider
+- CS3_CHUNK_SIZE - size of the downloaded fragment from Reva
+- CS3_SECURE_CHANNEL - secure channel flag
+- CS3_CLIENT_CERT - public key file path (PEM-encoded)
+- CS3_CLIENT_KEY - private key file path
+- CS3_CA_CERT - certificate authority file path
+```
```
michzimny, line 173: This list seems not to be complete. E.g. `CS3_HOME_DIR` or `CS3_LOGIN_TYPE` are missing. Not sure for others.

michzimny:
I added one small issue. I agree with all the previous comments from @diocas.

PR #38, opened 2020-11-30
Implemented config reading flow

mateusz-garlacz:
So the config priority ascending is: default values, config file, env vars. Host address and client id are required to run the plugin.
@diocas @michzimny 
#26 #27

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,8 +38,6 @@ RUN cd /opt/cs3 && \
 
 RUN fix-permissions "/home/${NB_USER}"
 
-ENV JUPYTER_ENABLE_LAB = 1
-
 EXPOSE 8888
 
 USER $NB_UID
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,8 @@ RUN cd /opt/cs3 && \
 
 RUN fix-permissions "/home/${NB_USER}"
 
+ENV JUPYTER_ENABLE_LAB = 1
+
 EXPOSE 8888
 
 USER $NB_UID
--- a/README.md
+++ b/README.md
@@ -221,7 +221,23 @@ Build docker image:
 ```bash
 docker build -t cs3api4lab .
 ```
-
+Available environmental variables:
+```
+- CS3_REVA_HOST - address and port on which the Reva server is listening [required]
+- CS3_CLIENT_ID - client login to authenticate in Reva [required]
+- CS3_CLIENT_SECRET - client password to authenticate in Reva [required in case of basic login]
+- CS3_AUTH_TOKEN_VALIDITY - the lifetime of the authenticating token
+- CS3_ENDPOINT - endpoint for Reva storage provider
+- CS3_CHUNK_SIZE - size of the downloaded fragment from Reva
+- CS3_SECURE_CHANNEL - secure channel flag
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -38,8 +38,6 @@ RUN cd /opt/cs3 && \
 
 RUN fix-permissions "/home/${NB_USER}"
 
-ENV JUPYTER_ENABLE_LAB = 1
```
michzimny, line 41: Why to remove this?

I guess that without this, the container would start with the classic UI, not jupyterlab. If we implement a plugin for jupyterlab, then it makes sense to run jupyterlab by default, and thus to keep this variable.

README.md
```diff
@@ -221,7 +221,23 @@ Build docker image:
...
+- CS3_CLIENT_ID - client login to authenticate in Reva [required]
+- CS3_CLIENT_SECRET - client password to authenticate in Reva [required in case of basic login]
+- CS3_AUTH_TOKEN_VALIDITY - the lifetime of the authenticating token
+- CS3_ENDPOINT - endpoint for Reva storage provider
+- CS3_CHUNK_SIZE - size of the downloaded fragment from Reva
+- CS3_SECURE_CHANNEL - secure channel flag
+- CS3_CLIENT_CERT - public key file path (PEM-encoded)
+- CS3_CLIENT_KEY - private key file path
+- CS3_CA_CERT - certificate authority file path
+```
```
michzimny, line 236: This list seems not to be complete. E.g. `CS3_HOME_DIR` or `CS3_LOGIN_TYPE` are missing. Not sure for others.

cs3api4lab/config/config_manager.py
```diff
@@ -1,54 +1,53 @@
...
+            # add self.config_dir to the front, if set manually
+            config_path.insert(0, os.getcwd() + self.__config_dir)
+        cm = ConfigManager(read_config_path=config_path)
+        config_file = cm.get(self.__config_file_name)
+        config = config_file.get("cs3")
+
+        # overwriting default values with config file
+        if config is not None:
+            for key in self.config.keys():
+                if key in config.keys() and config[key]:
```
michzimny, line 38: Why to check whether `config[key]` is not empty (i.e. the second condition)? In general, I think it's ok that one can override default config value with empty value (although there might be no use case in the current set of config variables).

Moreover, we do not check for emptiness in the case of env vars below, so it seems inconsistent.

Dockerfile
mateusz-garlacz: this var was commented before and I deleted it by mistake

michzimny: I'm not sure what do you mean by "before".

I see it is enabled in branch master. I think it's good to keep it enabled.

michzimny (approved):

michzimny:
I can't merge.

It says: 

> Merging is blocked
> The base branch restricts merging to authorized users. Learn more...

@diocas

diocas:
Merged. Next time, please automatically close the issues (by writing "closes #id" in the description).

I have one last comment (but not very important): to avoid doubts on why some config is not being taken, if would be nice to have better logging. Messages like "config XX overwritten by Env var" or "Unknown config XX " (for example, if some misspelled config was provided).

PR #41, opened 2020-12-14
Backend for access to share files and folders

JarCz:
Implements backend for access to share files by configuration: https://developer.sciencemesh.io/docs/iop/deployment/kubernetes/providers/


diocas (changes requested):
@JarCz please rebase. There are conflicts.

diocas (approved):

PR #43, opened 2021-01-07
Added User API endpoints

mateusz-garlacz:


diocas (approved):
@mateusz-garlacz I've approved and merged. 

Btw, if we have a swagger definition for sharing, why don't we have for the rest of the API?

mateusz-garlacz:
> @mateusz-garlacz I've approved and merged.
> 
> Btw, if we have a swagger definition for sharing, why don't we have for the rest of the API?

I am not sure if swagger is useful now, it was created to help Piotr develop frontend part but now we currently share information about the APIs internally in our team via rest client

PR #44, opened 2021-01-14
OCM API implementation

mateusz-garlacz:


cs3api4lab/api/cs3_ocm_share_api.py
```diff
@@ -0,0 +1,268 @@
...
+                "idp": share.owner.idp,
+                "opaque_id": share.owner.opaque_id
+            },
+            "creator": {
+                "idp": share.creator.idp,
+                "opaque_id": share.creator.opaque_id
+            }
+        }
+    
+    def _map_ocm_received_share(self, share):
```
diocas, line 225: Is it me, or this is exactly the same as `_map_ocm_share` with an extra `state`? I think we could avoid repetition..

diocas (commented):
General comment: you are dividing the log information in separate `self.log`, one with the message and the other with some other output. I would recommend keeping them together in a single `self.log`.

Then I have another question, more philosophical: should we have the ocm as a separate API? Couldn't we integrate it with the normal shares and hide the difference from the frontend?

I say this, but I would argue the frontend should always distinguish a normal share from an ocm one. Because I would do the following: the user searchs by an email, if the search returns no user, we should use it as an ocm share. But then I would put this as a parameter of the share api: "ocm = true or false". 
What do you think?

cs3api4lab/api/cs3_ocm_share_api.py
mateusz-garlacz: I fixed it

mateusz-garlacz:
> General comment: you are dividing the log information in separate `self.log`, one with the message and the other with some other output. I would recommend keeping them together in a single `self.log`.
> 
Fixed.
> Then I have another question, more philosophical: should we have the ocm as a separate API? Couldn't we integrate it with the normal shares and hide the difference from the frontend?
> 
> I say this, but I would argue the frontend should always distinguish a normal share from an ocm one. Because I would do the following: the user searchs by an email, if the search returns no user, we should use it as an ocm share. But then I would put this as a parameter of the share api: "ocm = true or false".
> What do you think?

I was thinking about the same. I agree that we could expose unified API, at least for normal shares and OCM shares. 
Don't forget about public shares, which probably have low priority now, but are also candidates to incorporate into unified API. I am not sure I understand how this search by emial mechanism would work, why not finding a user by emial would implicate ocm share?

diocas:
Yes, let's forget the public shares for now. Do you want to propose a unified API now or in a separate PR? 

About the email.. You need to know if you want to create OCM or normal, to know which cs3apis endpoint to contact right? This should be decided by the frontend (we could do in the backend, but this way we evoid double calls). When you search for a user you should be able to search by username or email. But, the results you get are only the ones that exist in your idp (this is correct, right?). So, if you are in your ailleron instance and search for "diogo@cern.ch", this will return no results. But, since it has the "@cern.ch", we know that this can be used as an OCM share, so the request to the sharing API should tell the backend that this is OCM. Does this make sense? I'm talking without much knowledge of the current OCM implementation, so I might have wrong assumptions...

---

Can you also do a overview of our API endpoints? A simple documentation would be fine, but if we could put the information as comments in the code and generate documentation based on it would be better. Can you open a issue?

mateusz-garlacz:
Yes, please merge this piece for clarity.
Now I get your idea. It makes sense, but I would like to look a bit deeper.
I will create an issue, code based documentation is preferable.

diocas (approved):
PR approved and merged. ToDo: #52 and #53

Issue #37, opened 2020-11-30
Access a shared folder


JarCz:
"Implement the way to navigate inside a shared folder, as if we were accessing one of our folders. We need to re-use the upstream component of listing files! Given this, I expect this to be relatively straight forward as well (unless you tell me that the components are not easily integrated). At this point, the shares API should resemble the normal files API and retrieve the same information, that's why I hope that using an upstream component would be straightforward."...

I tried to access a shared file or folder by modified Reva client with support storage id and file id. I created a test structure for "Einstein" user a shared it to "Marie".

We have a few things to discuss:

1. Path and content type for  https://cs3org.github.io/cs3apis/#cs3.sharing.collaboration.v1beta1.ListReceivedSharesResponse and https://cs3org.github.io/cs3apis/#cs3.sharing.collaboration.v1beta1.Share, 
now it hasn't a path and resource type. We choose away:
a. We use the "stat" command to determine this based on storage id and file id. We make an extra number of requests equals to the number of items in the received share.
b. cs3api extends the "Share" response on the path, file type, size, create time, based on https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.ResourceInfo

2. Mapping and hidden the virtual path in a file browser
For example, the file path "/home/share.txt" is mapping for Marie user as "/reva/einstein/share.txt". But path part "/reva/einstein" it's determinate from Reva config. I assume some Reva configuration can have a few different "share root folders".
Do we show all parts of the path? If yes, we don't need extra structure at the backend to map "file browser path" to "share path", but: 
* We present the Reva internal structure of shares. 
* It is a different folder structure for the user compared to a normal file structure.
* Probably we need to cache the response because jupyterlab reads the directory structure every few seconds

3. Hidden "MyShare" folder in the file browser.
In the file browser, we see the "MyShare". Do we need a list in the config to hidden some special folder and remove this item from the directory model? For me "MyShare" is a duplicate function like to "Share with me" tabs.

@diocas @michzimny what do you think about this?

PR #46, opened 2021-01-21
Add share path mapper with feature flag in the config

JarCz:
https://github.com/sciencemesh/cs3api4lab/issues/37 section 2

### DIFF
--- a/cs3api4lab/api/cs3_share_api.py
+++ b/cs3api4lab/api/cs3_share_api.py
@@ -19,6 +19,7 @@ import grpc
 
 from IPython.utils import tz
 
+from cs3api4lab.api.share_mapper import ShareMapper
 from cs3api4lab.auth import check_auth_interceptor
 from cs3api4lab.auth.authenticator import Auth
 from cs3api4lab.api.cs3_file_api import Cs3FileApi
@@ -36,6 +37,8 @@ class Cs3ShareApi:
     auth = None
     config = {}
     file_api = None
+    share_mapper = None
+
     date_fmt = '%Y-%m-%dT%H:%M:%SZ'
 
     TYPE_FILE = 1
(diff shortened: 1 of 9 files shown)


diocas:
Please rebase a add some description of what is being done in the PR. It's a bit confusing to understand at first sight.

JarCz:
Code in this PR cover the share path from IOP and transform into the normal relative path. If a user has the share path eg. "/reva/einsten/share_folder/share_test.txt" and on "Share by me" window click on the "/reva/ path can see other user data. (issue on IOP; https://github.com/cs3org/reva/issues/1390).  
The Mapping (share_mapper.py) creates an internal dictionary to store the original path and return to the frontend dictionary model with the mapped path. At reading containers or files it's executed the reverse process. If reading the path exists in a dictionary as the key it's replaced by the key-value and send to IOP instead of the path from the frontend.

diocas:
I still don't fully understand. Are you trying to avoid users from being able to click in the path that is gerated to access a share?
And I don't understand what the bug you found has to do with our lab extension. Once this gets fixed there's no problem for us, right?

Also if the shares are supposed to be accessible via our "Share" tab, we shouldn't expose "/reva/<user>". We should only see there "Shares/<some share id>/". This way the user will only be able to return to the shares dialog and nowhere else. (I said from the begining that we will have to change the breadcrumb eventually.)

And I definetely don't understand why you are putting an in memory cache. This shouldn't be needed for anything as the mappings you need to do should be done based on the share id and nothing else (meaning that the mappings are static).

---
At some point in the future I want us to discuss exposing the full namespace (since is what we will be doing at CERN with CERNBox). Meaning that the user will start (home) in the path "/eos/user/d/diogo" and the shares might point to this path as well. So take that in consideration.

JarCz:
I think we are able to freeze and close this PR now and will in the future, because:
- you are going to change the EOS names structure's
- yesterday presentation about changes in EOS 
- I spoke with Hugo on WP3 meet about this PR (https://github.com/cs3org/reva/issues/1390) and this bug should not have happened on a production system, because EOS watches over file access 

@diocas are you agree?

diocas:
Fine by me. But keep in mind that this extension needs to work in deployments without EOS as well. But the access control to other files will for sure be imposed in wtv storage backend is used.

Issue #11, opened 2020-09-03
Simple approach for concurrent opening of Notebooks


diocas:
I decided to move this issue from swan-cern/jupyter-extensions/issues/26 to this repo, since I believe it should be added to this extension. Some of the initial discussion can be checked in that issue.

---

While real-time collaborative editing of notebooks is a feature we would like to have, it would be useful to have an intermediate step, easier to accomplish, to allow users to open notebooks concurrently without causing conflicts.

This feature is of interest of JRC and CERN, and it would be useful for the project CS3MESH4EOSC. The current status of collaborative editing of notebooks is still uncertain, with a considerable effort still needed.

This ticket describes a possible approach, combining some of my ideas with the ones presented by JRC. Please let me know if you disagree with any of this.

## Approach

I suggest taking a similar approach to the editing of office files. This means, we would _lock_ a notebook if someone else is editing, which would only allow another person to open it in read-only mode. This would prevent the current Jupyter behavior of complaining that a file has changed on the disk and the possible loss off data.

But, we can still allow other's to contribute. We can ask the user if he just wants to open (ro) or if he wants to create a copy and merge it after. This copy should have a specific name, following some convention, that would allow the system to detect it and suggest a merge at a later stage.

## Implementation

To make this approach possible, we need to modify the upstream `Contents` interface. 
Opening a file should provide the following information:
* Is the notebook  _locked_ or not and by whom?;
* Do we have copies of the notebook?
* Do we have a copy of the notebook that was initiated by us?

We also need API endpoints to:
* Get the information about all the notebook copies (user that started, last save, is it ready to be merged or not)
* Set a notebook copy as "ready to merge";

This means that we need to create abstract methods for `locking`, `unlocking`, `getLockStatus`, `getConcurrentCopies`, `setReadyToMerge` etc.

We also need to change the UI to understand this new concept. It needs to understand that, in case a notebook is considered locked, it needs to ask the user to open in ro or create a copy. It also needs to suggest the merge of the changes. This can be done to the current user locking the file in another copy has been marked as "ready to merge", or to anyone else opening the file on which non merged copies still exist.

For the task of comparing notebooks, there's already an official tool called nbdime: https://nbdime.readthedocs.io/en/latest/

diocas:
@JarCz I propose we discuss this in the next meeting and maybe we could start working on some of the required components after you finish the authentication stuff and we check that the basic file access functionality is working.

mateusz-garlacz:
[...]

PR #56, opened 2021-03-07
Simple locking mechanism #11

mateusz-garlacz:
The idea is based on files metadata. The proposed flow is as follows:
- user creates a share with an original file
- the original file will contain in it's metadata information about locks i.e. who has a copy and if it is used or not
- when opening a file, check if it is a copy or original
- if you open a copy, the correspondent lock present in the original file's metadata will be updated
- if you open the original file, a copy and a lock will be created
- when closing a file, the endpoint deleting a lock should be invoked
- the contributor who has the merging rights would be the one who has the oldest and valid lock

Probably there are corrections to be done to cover some edge cases, but this is the idea in the nutshell. Please take a look and leave comments.

### DIFF
--- a/cs3api4lab/api/cs3_file_api.py
+++ b/cs3api4lab/api/cs3_file_api.py
@@ -21,6 +21,8 @@ from cs3api4lab.auth import check_auth_interceptor
 from cs3api4lab.auth.authenticator import Auth
 from cs3api4lab.auth.channel_connector import ChannelConnector
 from cs3api4lab.config.config_manager import Cs3ConfigManager
+from cs3api4lab.api.cs3_locks_api import LocksApi
+from cs3api4lab.api.storage_api import StorageApi
 
 
 class Cs3FileApi:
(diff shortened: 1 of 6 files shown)


diocas:
This is a long and complex PR, so I will need some time to review it. But can you give more details of the implementation? 
So, the lock information is only contained inside the file metadata? From a quick look, I got the impression that you kept some file(s) inside the user home dir? Why don't you create a hidden lock file, in the same directory and with its named based on the original file, just like the Office applications do?

mateusz-garlacz:
Probably I introduced some unclarity because of naming.
Reva is the source of truth for the information about who has a copy of a given file and if someone is currently working on corresponding copy. The information is stored in the metadata. in this context I used 'copy' and 'lock' interchangeably. The Office-like 'lock file' you are reffering to would be something I call a copy. 

When opening a shared file, a copy is created (how to unable editing of the original file and switch the view to another tab from the frontend perspective is another topic). A possible improvements I can already see is the copy is now named with user-opaque-idp pattern but can be changed to '$originalName' like in Office.

diocas:
@mateusz-garlacz I was going to review the PR again right now.. I guess you're going to replace it with something new? xD

mateusz-garlacz:
Well... I decided that after so long time it needs some refreshment, so stay tuned :D

PR #57, opened 2021-03-12
Tabbed file browser layout

JarCz:


diocas (approved):
Not going to review in detail. For now, I think it's better that we merge and test, and then we improve (clean unnecessary stuff, organize, etc).

PR #58, opened 2021-03-23
Fix cs3apis 54dev

JarCz:


diocas (approved):

Dockerfile
```diff
@@ -9,7 +11,6 @@ COPY jupyter-config/cs3api4lab.json /opt/cs3/jupyter-config/cs3api4lab.json
 COPY setup.py /opt/cs3/setup.py
 COPY README.md /opt/cs3/README.md
 COPY package.json /opt/cs3/package.json
-COPY package-lock.json /opt/cs3/package-lock.json
```
diocas, line 12: why is this a problem?

JarCz: It's a small optimization because npm automatically generates the "package-lock" file for any operations. It's better if we don't copy from the source code.

diocas: NPM generates a new package-lock file, yes, but the idea of a lock file is to lock versions to avoid building with dependencies that have not been tested. So the recommendation is that we **never** delete this file.
This file should be managed by us and versions should only be updated when we say so. By removing it, we might break compilation in the future, as this will always pull the latest versions of the dependencies.

JarCz: 👍

diocas (approved):

PR #62, opened 2021-05-31
Upgraded plugin to Jupyter 3

mateusz-garlacz:


README.md
```diff
@@ -27,10 +27,10 @@ jupyter serverextension enable --py cs3api4lab --sys-prefix
 jupyter labextension install @sciencemesh/cs3api4lab
 ```
 
-To enable the Manager, the following configuration needs to be added to `jupyter_notebook_config.py`:
+To enable the Manager, the following configuration needs to be added to `jupyter_server_config.py`:
 
 ```python
-c.NotebookApp.contents_manager_class = 'cs3api4lab.CS3APIsManager'
+c.ServerApp.contents_manager_class = 'cs3api4lab.api.cs3apismanager.CS3APIsManager'
```
diocas, line 33: Why did we remove `from cs3api4lab.api.cs3apismanager import CS3APIsManager` from `__init__`? Then we could still use the simplified version instead.

README.md
```diff
@@ -121,20 +115,20 @@ python test_cs3apismanager.py
...
 ```
-c.NotebookApp.contents_manager_class = 'notebook.services.contents.largefilemanager.LargeFileManager'
+c.ServerApp.contents_manager_class = 'notebook.services.contents.largefilemanager.LargeFileManager'
 ```
 
 to
 
 ```
-c.NotebookApp.contents_manager_class = 'cs3api4lab.CS3APIsManager'
+c.ServerApp.contents_manager_class = 'cs3api4lab.api.cs3apismanager.CS3APIsManager'
```
diocas, line 131: Same as above

cs3api4lab/__init__.py
```diff
@@ -1,19 +1,35 @@
-from notebook.utils import url_path_join
+import json
+from pathlib import Path
 
+from .handlers import setup_handlers
 from ._version import __version__
-from cs3api4lab.api.cs3apismanager import CS3APIsManager
```
diocas, line 4: Already asked above. By removing this then we need to have a not as clean configuration. Not a big issue, but I'm curious to know why we had to remove it.

cs3api4lab/api/cs3_file_api.py
```diff
@@ -68,7 +68,18 @@ def stat(self, file_id, endpoint=None):
             }
 
         self.log.info('msg="Failed stat" fileid="%s" reason="%s"' % (file_id, stat_info.status.message))
-        raise FileNotFoundError(stat_info.status.message + ", file " + file_id)
+        return {
```
diocas, line 71: Does this cause an issue in the frontend?

cs3api4lab/handlers.py
```diff
@@ -262,20 +274,27 @@ def get(self):
...
+        (r"/api/cs3/ocm/received", OcmReceivedSharesHandler),
+        (r"/api/cs3/user", UserInfoHandler),
+        (r"/api/cs3/user/claim", UserInfoClaimHandler),
+        (r"/api/cs3/user/query", UserQueryHandler)
+    ]
+
+    for handler in handlers:
+        pattern = url_path_join(web_app.settings['base_url'], handler[0])
+        new_handler = tuple([pattern] + list(handler[1:]))
+        print("ADDING HANDLER " + str(handler))
```
diocas, line 296: Log and not print

setup.py
```diff
@@ -1,55 +1,88 @@
...
+    author=pkg_json["author"],
+    description=pkg_json["description"],
+    license=pkg_json["license"],
+    long_description=long_description,
     long_description_content_type="text/markdown",
+    cmdclass=cmdclass,
     packages=setuptools.find_packages(),
     install_requires=[
-        "jupyterlab~=2.0",
+        "jupyterlab==3.0.16",
```
diocas, line 69: Does it need to be this specifically? If there's nothing particularly blocking, I would allow some margin, at least on patch version, maybe minor as well.

diocas (commented):
@mateusz-garlacz thanks for the PR. Just some small comments and we can merge.

By looking at it I was reminded that we never had the PR to combine the sharing API, right?

cs3api4lab/handlers.py
```diff
@@ -1,14 +1,26 @@
+import os
+import json
+
+from notebook.base.handlers import APIHandler
```
diocas, line 4: There's another thing that I forgot to mention.. You made this extension compatible with jupyter-server. But then you are using `notebook` and retrieving its methods+classes. While this should be ok for now, it might not be in the future so we could take the opportunity to improve the dependencies. I would remove notebook from the dependencies completely.

Dockerfile
```diff
@@ -20,25 +15,21 @@ USER root
...
-    sed -i -e '$ac.NotebookApp.contents_manager_class = \x27cs3api4lab.CS3APIsManager\x27' /etc/jupyter/jupyter_notebook_config.py
-
-RUN fix-permissions "/home/${NB_USER}"
+    jupyter lab clean -y && \
+    npm cache clean --force && \
+    jupyter server --generate-config -y && \
+    jupyter lab --generate-config -y && \
+    rm -rf "/home/${NB_USER}/.cache/yarn" && \
+    rm -rf "/home/${NB_USER}/.node-gyp" && \
+    sed -i -e '$ac.NotebookApp.contents_manager_class = \x27cs3api4lab.api.cs3apismanager.CS3APIsManager\x27' /etc/jupyter/jupyter_notebook_config.py && \
```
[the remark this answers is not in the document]
mateusz-garlacz: Actually is seems as if the image used Notebook instead Server, it does not pick up jupyter_server config for some reason

README.md
mateusz-garlacz: The upgrade script messed up init.py a bit and I forgot about this trick. Fixed.

cs3api4lab/api/cs3_file_api.py
mateusz-garlacz: Yes, it prevented shares from rendering

cs3api4lab/handlers.py
mateusz-garlacz: Sure we can do that, but I am a bit confused because the examples https://github.com/jupyterlab/extension-examples are Jupyter 3 compatible and still use notebook imports

mateusz-garlacz:
Hi, right, the PR with API unification hasn't been created or withdrawn yet, it will come as second after this one is merged

cs3api4lab/api/cs3_file_api.py
diocas: And not having a return of 404 doesn't break anything else?

cs3api4lab/handlers.py
diocas: Let's create a ticket and do that after, then, if ok with you.

Dockerfile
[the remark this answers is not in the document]
diocas: uhm, strange.. it might be because it's still using notebook in this image, I don't know. Let's keep it for now, but we should fix it.

diocas:
@mateusz-garlacz Just one last round of questions/request to create an issue, and we can merge.

cs3api4lab/api/cs3_file_api.py
mateusz-garlacz: No, it just ommits the empty share when rendering created/received shares. I'll come up with something more clear with Api unification PR. This is part of a bigger question, i. e. how to handle moving/renaming/deleting shared files, because shares are indepentent entities from the files. I need to think about it. But let's focus on this now.

cs3api4lab/handlers.py
mateusz-garlacz: https://github.com/sciencemesh/cs3api4lab/issues/63

PR #65, opened 2021-06-24
search_user

piotrWichlinskiSoftwaremind:
change user selection box to use endpoint filters for searching user

### DIFF
--- a/src/createShare.tsx
+++ b/src/createShare.tsx
@@ -1,5 +1,5 @@
 import {ReactWidget} from '@jupyterlab/apputils';
-import React, {useEffect, useState} from 'react';
+import React, { useState} from 'react';
 import {Contents} from '@jupyterlab/services';
 import {requestAPI} from './services';
 import {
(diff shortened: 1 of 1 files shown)


src/createShare.tsx
```diff
@@ -46,19 +46,18 @@ const ShareForm: React.FC<ShareFormProps> = (
 
     const [formValues, setFormState] = useState({
         endpoint: '/',
-        file_path: '/home/' + props.fileInfo.path.replace('cs3drive:', ''),
+        file_path: '/home/' + shareProps.fileInfo.path.replace('cs3drive:', ''),
```
diocas, line 49: I'm pretty sure we should not hardcode `/home`. At CERN we still have it for now but don't use it, so I don't know if we should make this configurable. Can you check?

diocas: But since this was already like that, I'm approving+merging and we change later.

diocas (approved):

PR #68, opened 2021-07-22
Enabled hidden files

mateusz-garlacz:
This PR is to fix hidden files. Hiding files starting with a dot was implemented sometime in the begining of development but didn't work. This will be also useful in locks.


diocas (changes requested):
Can we make the show/hide hidden files configurable?
It was actually a functionality that is part of the frontend mockups: https://user-images.githubusercontent.com/2589871/87484282-1620bf80-c636-11ea-9122-59092b0011c7.png

diocas:
Or, since in that mockup the frontend knows how many are hidden, I guess this should be functionality for the frontend, making this PR unecessary.

PR #72, opened 2021-08-30
Bottom widget

mateusz-garlacz:
Hi, this is the enhancement for bottom widget for covering/discovering hidden files as discussed in #68

### DIFF
--- a/src/cs3panel.tsx
+++ b/src/cs3panel.tsx
@@ -93,16 +93,13 @@ export class Cs3HeaderWidget extends ReactWidget {
 
 export const Bottom = (props: BottomProps): JSX.Element => {
 
-    const [text, setText] = useState(props.message);
+    const [text, setText] = useState('Show hidden files');
 
     const setLabel = async () => {
         const showHidden: boolean = await props.db.fetch('showHidden') as boolean;
-        if (showHidden == undefined || !showHidden) {
-            const hiddenFilesNo: number = await props.db.fetch('hiddenFilesNo') as number;
-            setText('Hidden files ' + hiddenFilesNo + ' (show)')
-        } else {
-            setText('Cover hidden files')
-        }
+        const hiddenFilesNo: number = await props.db.fetch('hiddenFilesNo') as number;
+        const action = (showHidden == undefined || !showHidden) ? 'show' : 'hide'
+        setText(`${hiddenFilesNo} hidden files (${action})`)
     }
 
     props.browser.model.pathChanged.connect( async (browser, args) => {
@@ -123,7 +120,7 @@ export const Bottom = (props: BottomProps): JSX.Element => {
 }
 
 export class Cs3BottomWidget extends ReactWidget {
-    private bottomProps: { message: string; db: IStateDB; drive: CS3Contents, browser: FileBrowser };
+    private bottomProps: { db: IStateDB; drive: CS3Contents, browser: FileBrowser };
 
     constructor(title: string, id: string, options: Widget.IOptions = {},
                 stateDB: IStateDB,
@@ -133,7 +130,7 @@ export class Cs3BottomWidget extends ReactWidget {
         this.addClass('c3-bottom-widget');
         this.id = id;
         this.title.closable = false;
-        this.bottomProps = {message: 'Show hidden files', db: stateDB, drive: drive, browser: browser}
+        this.bottomProps = {db: stateDB, drive: drive, browser: browser}
         this.node.onclick = async () => {
             const showHidden = await stateDB.fetch('showHidden')
             await stateDB.save('showHidden', !showHidden);
@@ -142,9 +139,8 @@ export class Cs3BottomWidget extends ReactWidget {
     }
 
     protected render(): JSX.Element {
-        return <Bottom message={this.bottomProps.message}
+        return <Bottom
                        db={this.bottomProps.db}
-                       drive={this.bottomProps.drive}
                        browser={this.bottomProps.browser}/>;
     }
 }
--- a/src/cs3panel.tsx
+++ b/src/cs3panel.tsx
@@ -7,8 +7,6 @@ import {BottomProps} from "./types";
 import {useState} from "react";
 import {FileBrowser} from "@jupyterlab/filebrowser";
 import {CS3Contents} from "./drive";
-import {Contents} from "@jupyterlab/services";
-import {requestAPI} from "./services";
 
 export class Cs3Panel extends Widget {
     protected header: BoxPanel;
@@ -97,33 +95,27 @@ export const Bottom = (props: BottomProps): JSX.Element => {
 
     const [text, setText] = useState(props.message);
 
-    const onPathOrFileChanged = async (path: string) => {
+    const setLabel = async () => {
         const showHidden: boolean = await props.db.fetch('showHidden') as boolean;
         if (showHidden == undefined || !showHidden) {
-            const result: Contents.IModel = await requestAPI('/api/contents/' + path, {method: 'get'});
-            const hiddenFilesNo = result.content.filter((file: { name: string; }) => file.name.startsWith('.')).length
+            const hiddenFilesNo: number = await props.db.fetch('hiddenFilesNo') as number;
             setText('Hidden files ' + hiddenFilesNo + ' (show)')
         } else {
             setText('Cover hidden files')
         }
     }
 
-    props.browser.model.pathChanged.connect(async (browser, args) => {
-        onPathOrFileChanged(args.newValue.replace('cs3drive:', ''))
+    props.browser.model.pathChanged.connect( async (browser, args) => {
+        await setLabel()
     })
 
-    props.browser.model.refreshed.connect((browser) => {
-        onPathOrFileChanged(browser.path.replace('cs3drive:', ''))
+    props.browser.model.refreshed.connect(async (browser) => {
+        await setLabel()
     })
 
     return (
         <div className={'jp-bottom-div'} onClick={async () => {
-            if (text.startsWith('Hidden files') || text.startsWith('Show hidden')) {
-                setText('Cover hidden files')
-            } else {
-                const hiddenFilesNo = await props.db.fetch('hiddenFilesNo')
-                setText('Hidden files ' + hiddenFilesNo + ' (show)')
-            }
+            await setLabel()
         }}>
             {text}
         </div>
--- a/src/cs3panel.tsx
+++ b/src/cs3panel.tsx
@@ -111,11 +111,7 @@ export const Bottom = (props: BottomProps): JSX.Element => {
     })
 
     return (
-        <div className={'jp-bottom-div'} onClick={async () => {
-            await setLabel()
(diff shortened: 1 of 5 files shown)


src/cs3panel.tsx
```diff
@@ -89,18 +93,67 @@ export class Cs3HeaderWidget extends ReactWidget {
...
 }
 
+export const Bottom = (props: BottomProps): JSX.Element => {
+
+    const [text, setText] = useState(props.message);
+
+    const onPathOrFileChanged = async (path: string) => {
+        const showHidden: boolean = await props.db.fetch('showHidden') as boolean;
+        if (showHidden == undefined || !showHidden) {
+            const result: Contents.IModel = await requestAPI('/api/contents/' + path, {method: 'get'});
```
diocas, line 103: So you do an extra call (repeated), just to count the current hidden files? Isn't there a way to get this info when we receive the current folder listing? (didn't you do that drive.ts already?)
I would expect that showing/hiding hidden files would result in 0 calls to the server. Ever.

src/cs3panel.tsx
```diff
@@ -89,18 +93,67 @@ export class Cs3HeaderWidget extends ReactWidget {
...
+            const hiddenFilesNo = result.content.filter((file: { name: string; }) => file.name.startsWith('.')).length
+            setText('Hidden files ' + hiddenFilesNo + ' (show)')
+        } else {
+            setText('Cover hidden files')
+        }
+    }
+
+    props.browser.model.pathChanged.connect(async (browser, args) => {
```
diocas, line 111: Not knowledgeable enough on lab inner workings, but don't we have a way to subscribe to contents changed and fetch the number information from there?

src/cs3panel.tsx
```diff
@@ -89,18 +93,67 @@ export class Cs3HeaderWidget extends ReactWidget {
...
+        onPathOrFileChanged(args.newValue.replace('cs3drive:', ''))
+    })
+
+    props.browser.model.refreshed.connect((browser) => {
+        onPathOrFileChanged(browser.path.replace('cs3drive:', ''))
+    })
+
+    return (
+        <div className={'jp-bottom-div'} onClick={async () => {
+            if (text.startsWith('Hidden files') || text.startsWith('Show hidden')) {
```
diocas, line 121: Don't understand the repetition of this set text nor what we're checking. Shouldn't instead have a boolean "props.showHidden" (or smth) i.e ==true when showing, which sets the corresponding text in a single place?

src/index.ts
```diff
@@ -182,32 +187,32 @@ const cs3browser: JupyterFrontEndPlugin<void> = {
         const cs3Panel = new Cs3Panel('cs3 panel', 'cs3-panel', kernelIcon);
 
         stateDB.save('share', {share_type: 'filelist'});
+        stateDB.save('showHidden', false);
```
diocas, line 190: What is this stateDB? does it persist in the browser localstorage? If so, doesn't this reset the config everytime the user loads the page?

src/drive.ts
```diff
@@ -230,9 +225,15 @@ async function getFileList(
...
 
-    return await requestAPI('/api/contents/' + path + '' + url, {
-        method: 'get'
-    });
+    const result: Contents.IModel  = await requestAPI('/api/contents/' + path + '' + url, {method: 'get'});
+    if (!showHidden && Array.isArray(result.content)) {
+        const hiddenFilesNo: number = result.content.filter((file: { name: string; }) => file.name.startsWith('.')).length
+        await stateDB.save('hiddenFilesNo', hiddenFilesNo)
+        const filteredResult = JSON.parse(JSON.stringify(result))
+        filteredResult.content = (result.content as Array<any>).filter((file: { name: string; }) => !file.name.startsWith('.'))
```
diocas, line 233: I mention that maybe we should not do any request to backend when changing between show/hide. So, can't we store this somehow and filter only when printing in the UI?

diocas (commented):
Can you put a screenshot of this to have a better idea of how it looks?

I confess I'm not knowledgeable in these technologies, but by looking at this PR I see so much duplicated/repeated stuff that I wonder if it really needs to be like this. Text repeated, duplicated calls to backed or multiple places requesting the same info, same variables passed through many layers. Can't we keep all of this number of hidden files + isHidden or shown variables in the files model and use it in the graphical elements (by subscribing to changes on those variables)? 

I might be wrong, as I said, but this seems very complicated to maintain, which is my biggest worry. Not only for this PR but for the extension in general.

src/cs3panel.tsx
mateusz-garlacz: hidden-hide
hidden-show

src/cs3panel.tsx
mateusz-garlacz: Unfortunately browser.model events don't expose info about current content, or the access is restricted

src/index.ts
mateusz-garlacz: stateDB is a handy storage to share info between components and yes, after refresh the data is destroyed, but after refresh it's loaded again when jl pulls file list

src/drive.ts
mateusz-garlacz: I simplified it, now there are no additional calls to backend, the information about hidden files number is fetched during file list request

src/index.ts
diocas: Should we make it persistent in localstorage?

src/cs3panel.tsx
```diff
@@ -89,18 +91,61 @@ export class Cs3HeaderWidget extends ReactWidget {
...
 
+export const Bottom = (props: BottomProps): JSX.Element => {
+
+    const [text, setText] = useState(props.message);
+
+    const setLabel = async () => {
+        const showHidden: boolean = await props.db.fetch('showHidden') as boolean;
+        if (showHidden == undefined || !showHidden) {
+            const hiddenFilesNo: number = await props.db.fetch('hiddenFilesNo') as number;
+            setText('Hidden files ' + hiddenFilesNo + ' (show)')
```
diocas, line 102: X hidden files (show)

src/cs3panel.tsx
```diff
@@ -89,18 +91,61 @@ export class Cs3HeaderWidget extends ReactWidget {
...
+        } else {
+            setText('Cover hidden files')
```
diocas, line 104: X hidden files (hide) ?

src/cs3panel.tsx
```diff
@@ -89,18 +91,61 @@ export class Cs3HeaderWidget extends ReactWidget {
...
+    props.browser.model.pathChanged.connect( async (browser, args) => {
+        await setLabel()
+    })
+
+    props.browser.model.refreshed.connect(async (browser) => {
+        await setLabel()
+    })
+
+    return (
+        <div className={'jp-bottom-div'} onClick={async () => {
```
diocas, line 117: Do we even change `showHidden` to true or false (I'm only seeing fetch)? SHouldn't we be doing it here?

src/cs3panel.tsx
```diff
@@ -89,18 +91,61 @@ export class Cs3HeaderWidget extends ReactWidget {
     }
 }
 
+export const Bottom = (props: BottomProps): JSX.Element => {
+
+    const [text, setText] = useState(props.message);
```
diocas, line 96: Can't we initiate with empty string and remove this props.message?

src/cs3panel.tsx
```diff
@@ -89,18 +91,61 @@ export class Cs3HeaderWidget extends ReactWidget {
...
+            await setLabel()
+        }}>
+            {text}
+        </div>
+    )
+}
+
 export class Cs3BottomWidget extends ReactWidget {
-    constructor(title: string, id: string, options: Widget.IOptions = {}) {
+    private bottomProps: { message: string; db: IStateDB; drive: CS3Contents, browser: FileBrowser };
```
diocas, line 126: do we need message (see above) and drive?

src/drive.ts
diocas: But we still get to the backend every time we switch between show/hide the hidden files, right?

diocas (changes requested):
Looks much better, thanks!
But I have some comments and questions.

src/index.ts
mateusz-garlacz: Actually after double check the data persisted in stateDB is not affected by refreshments, I'm pretty sure it uses localstorage underneath. This is the default way to store vars in jupyter, I don't think we need to switch to localstorage explicitly.

src/drive.ts
mateusz-garlacz: No, Jupyter pings /contents endpoint every several seconds by default and pings this endpoint as well during directory change or file list refresh, I use result of this request to count current number of files. When switching between show/hide no additional request is made, it's just for label change.

src/cs3panel.tsx
mateusz-garlacz: I can remove message from props, but if we initiate with empty string there will be no text on the bar after start

src/cs3panel.tsx
```diff
@@ -89,18 +91,61 @@ export class Cs3HeaderWidget extends ReactWidget {
...
+                drive: CS3Contents) {
         super(options);
         this.addClass('c3-bottom-widget');
         this.id = id;
-        this.title.label = title;
-        this.title.caption = title;
         this.title.closable = false;
+        this.bottomProps = {message: 'Show hidden files', db: stateDB, drive: drive, browser: browser}
+        this.node.onclick = async () => {
+            const showHidden = await stateDB.fetch('showHidden')
```
mateusz-garlacz, line 138: 'showHidden' is toggled here

src/cs3panel.tsx
mateusz-garlacz: It seems we don't

src/cs3panel.tsx
diocas: so why don't we do the `setLabel` here as well? It would be easier to follow as we have a single `onclick` event (I'm assuming that `node` is the `div`, which I'm not sure).

diocas: But, the fact that you `refresh` the model shouldn't already be changing the text automatically (since you listen for model change)?

src/drive.ts
diocas: ok!

src/index.ts
diocas: perfect!

diocas (commented):
Just one little comment more and we merge.

src/cs3panel.tsx
mateusz-garlacz: So this component does not have access to setLabel and it cannot change Bottom component's state, instead it uses onClick to trigger browser's model refreshment, which in turn triggers setting label

mateusz-garlacz: Btw, would you like to squash commits into single one?

diocas: Ok, so my second comment made sense. Perfect. I can squash when merging.

diocas (approved):
Ready to merge

PR #88, opened 2021-12-03
Combine modals

piotrWichlinskiSoftwaremind:
This contains strict null check from the previous PR, so it should be merged after strict null checks


piotrWichlinskiSoftwaremind:
This changes will be in combine_modals_v2

PR #89, opened 2021-12-06
Eslint

piotrWichlinskiSoftwaremind:
These are eslint fixes, it should be merged after modal_combine PR to avoid problems. This PR contains tab size changes, return types warnings jsdoc updates etc.


diocas:
You're applying linting here instead of the individual PRs you opened? If so, I can overlook them for now, but it would've been better to do it on each PR.

piotrWichlinskiSoftwaremind:
This was done in a separate branch because of the number of changes it contains. Every task that we are working on from this time will contain linter fixes.

diocas:
@piotrWichlinskiSoftwaremind still failing the ci, but at least now I can merge. You also have the combined modal commit in the middle of a linting PR, but I can merge if you don't want to fix it.

piotrWichlinskiSoftwaremind:
@diocas These PRs are chained, each one contains previous, so if you merge it in the correct order it should be fine. This is the last one that contains previous ones (strict values check and combined modals) and fixes the linter for all of them.

piotrWichlinskiSoftwaremind:
This change has been included in strict_null_checks and combine_modals so there is no need for this branch/PR

PR #92, opened 2022-01-17
Strict null check

diocas:


diocas:
@piotrWichlinskiSoftwaremind @mateusz-garlacz @dagl I've merged Piotrs PR with some fixes to make the CI pass (I've had to disable testing for now, we will have to implement that later). Please rebase all other PRs.

PR #93, opened 2022-01-18
combined modals with eslint/prettier fixes

piotrWichlinskiSoftwaremind:


src/infobox.tsx
```diff
@@ -351,3 +283,151 @@ const Shares = (props: SharesProps): JSX.Element => {
...
+ *
+ * @constructor
+ */
+const CreateShare = (props: CreateShareProps): JSX.Element => {
+  return (
+    <>
+      <ShareForm
+        fileInfo={props.fileInfo}
+        getUsers={async (query): Promise<Array<UsersRequest>> => {
+          return await requestAPI('api/cs3/user/query?query=' + query, {});
```
diocas, line 414: `api/` vs `/api/` bellow. We use `join`, so it doesn't matter, but at least would be more consistent.

diocas (commented):
Just 2 comments. Otherwise let's merge.

diocas:
@piotrWichlinskiSoftwaremind only saw your messages now, sorry. Next time please ping here in the conversation as well so that I receive an email.

I was wondering if we really need to return the mountpoint to the frontend... Can't we assume that every request is relative to the mountpoint (meaning that we would combine both, the mountpoint and the path, in the backend)? Or would this break anything? or we will allow users to go up in tree, inclusive to other mountpoints?

If not, we should remove all of these details from the frontend and try to simplify as much as possible.

diocas (approved):
The fix to remove the home request will be followed in #99

PR #94, opened 2022-01-20
Fixed opening file

mateusz-garlacz:
NOTE: this should be merged after combined modals changes

### DIFF
--- a/src/drive.ts
+++ b/src/drive.ts
@@ -232,18 +232,19 @@ async function getFileList(
     '/api/contents/' + path + '' + url,
     { method: 'get' }
   );
+  if (path !== null && !path.includes('.')) {
+    const hiddenFilesNo: number = result.content.filter(
+      (file: { name: string }) => file.name.startsWith('.')
+    ).length;
+    await stateDB.save('hiddenFilesNo', hiddenFilesNo);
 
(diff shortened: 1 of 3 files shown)


src/drive.ts
```diff
@@ -232,18 +232,19 @@ async function getFileList(
...
-    );
-    return filteredResult;
+  if (path !== null && !path.includes('.')) {
+    const hiddenFilesNo: number = result.content.filter(
+      (file: { name: string }) => file.name.startsWith('.')
+    ).length;
+    await stateDB.save('hiddenFilesNo', hiddenFilesNo);
+
+    if (!showHidden && Array.isArray(result.content)) {
+      const filteredResult = JSON.parse(JSON.stringify(result));
```
diocas, line 242: Why are we stringify and parse at the same time? If there's a type issue, can't you cast?

diocas (changes requested):

src/drive.ts
mateusz-garlacz: This is a workaround to clone the object in order to substitute the list of files. Casting to a type won't help, since Contents.IModel is immutable

diocas (approved):

PR #97, opened 2022-01-20
lgtm alert fixes for frontend

piotrWichlinskiSoftwaremind:


src/infobox.tsx
```diff
@@ -34,10 +33,6 @@ class Main extends React.Component<any, any> {
    */
   public constructor(props: MainProps) {
```
diocas, line 34: Maybe we can remove the constructor completely, no?

piotrWichlinskiSoftwaremind: @diocas we can't remove it, I need it to inject file information into react props

diocas: But isn't the parent constructor already receiving that information? Usually you only redefine a method if you need to add extra logic, but here you just call the parent.

PR #104, opened 2022-02-04
fixed packages version // small lint fixes

piotrWichlinskiSoftwaremind:
Packages versions have to match exactly versions of the installed jupyter


diocas:
Didn't we agree to keep Lab 3.0 for now? In SWAN I cannot upgrade.. Or are these versions backwards compatible (I guess no?)?

piotrWichlinskiSoftwaremind:
These are package versions of the 3.0, previous package versions were incorrect, it installed the newest.

diocas:
Ah yea, I saw the wrong thing.. Merging!

PR #111, opened 2022-02-16
fix config for other platforms

dagl:
A quick fix, because config couldn't be read on Linux


diocas (approved):
Ok, either way, being this Jupyter, the configuration paths are already well know and do work. So I accept this path for testing, but it should be discouraged.

PR #112, opened 2022-02-16
Made OCM optional and temporarily disabled

mateusz-garlacz:


cs3api4lab/api/share_api_facade.py
```diff
@@ -39,7 +39,10 @@ def __init__(self, log):
     def create(self, endpoint, file_path, opaque_id, idp, role=Role.EDITOR, grantee_type=Grantee.USER, reshare=True):
         """Creates a share or creates an OCM share if the user is not found in local domain"""
         if self._is_ocm_user(opaque_id, idp):
-            return self.ocm_share_api.create(opaque_id, idp, idp, endpoint, file_path, grantee_type, role, reshare)
+            if self.config['enable_ocm']:
+                return self.ocm_share_api.create(opaque_id, idp, idp, endpoint, file_path, grantee_type, role, reshare)
+            else:
```
diocas, line 44: We should fail as this request should not reach us. `501`? 
We should do this for all other cases where it applies such an error.

diocas (changes requested):
Please add errors instead of just silently failing. The frontend needs to be aware this is optional and handle attempts from users of sharing with OCM.
Please also rebase.

dagl:
I think we should extend this functionality project-wise, for example in cs3_user_api:find_user_by_query we also try to use ocm

diocas:
You also try OCM when searching for a users, but you only if it's enabled. I was expecting this PR to fix all those cases (the web part, maybe can come in a 2nd PR).

mateusz-garlacz:
> You can also try OCM when searching for a users, but you only if it's enabled. I was expecting this PR to fix all those cases (the web part, maybe can come in a 2nd PR).

I updated the PR. What do you mean by web part ?

cs3api4lab/api/cs3_user_api.py
```diff
@@ -50,12 +50,13 @@ def find_users_by_query(self, query):
         ocm_response = self.invite_api.FindAcceptedUsers(ia.FindAcceptedUsersRequest(filter=query),
```
diocas, line 50: Shouldn't this be inside the if as well?

diocas (commented):
Just one last comment and we can merge.

What I meant about the web is that it should not receive the errors you just added. Those are for anyone trying to circumvent the API. The web should prevent any attempt to use OCM if it's not available. I don't know if that's already the case, but I would expect it to be (for example, if searching for a user the OCM options do not appear, it's expectable that they are not used when creating a share so all is good). So this was just a remark :)

diocas:
@mateusz-garlacz when you removed the ocm handlers you forgot to remove the import and LGTM is now complaining. (and thanks for removing the handlers, now we really have a unified api!)

mateusz-garlacz:
Ah sorry, my bad. Now I made lgtm happy

PR #114, opened 2022-02-24
Remove default filebrowser

piotrWichlinskiSoftwaremind:
- disable standard file browser 
- include some code from the filebrowser-plugin 
- update docker file to add an additional config file 
- add page_config 
- update readme
- fix file paths for new files

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,17 @@ to
 ```
 c.ServerApp.contents_manager_class = 'cs3api4lab.CS3APIsManager'
 ```
+### Disable default file browser
+To disable the default file browser use these commands in the console:
+```bash
+jupyter labextension disable @jupyterlab/filebrowser-extension
+jupyter labextension disable @jupyterlab/filebrowser
+```
+You can also use a config file to disable the default file browser, copy page_config.json from jupyter-config/page_config.json to:
+* Windows:
+  ```C:\Users\{USER_PROFILE}\.jupyter\labconfig```
+* Linux:
+  ```HOME_FOLDER/.jupyter/labconfig```
 
 ### CS3 config file
 Copy cs3 example config file from "jupyter-config/jupyter_cs3_config.json"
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -154,6 +154,12 @@ to
 ```
 c.ServerApp.contents_manager_class = 'cs3api4lab.CS3APIsManager'
 ```
+### Disable default file browser
+Copy page_config.json from jupyter-config/page_config.json to:
```
diocas, line 158: I would suggest `jupyter labextension disable` as an easier way to enable/disable extensions. Maybe could consider that to the docker image as well.

src/browserCommands.ts
```diff
@@ -0,0 +1,880 @@
+/**
+ * The command IDs used by the file browser plugin.
```
diocas, line 2: This is a complete copy of upstream, with very small changes, right?
So please add that information in the comment at the top. If there are many changes, please add comments on the specific places detailing what's different (to make it easier to update in the future)

diocas (changes requested):

README.md
piotrWichlinskiSoftwaremind: Of course, this is an option too, I have updated a readme to add this information and added commands for dockerfile instead of config

src/browserCommands.ts
piotrWichlinskiSoftwaremind: It's not a full copy, I have copied what we needed and do some small changes, mostly exports so it could be in a separate file, also changed the paths for file creation. I added a comments

diocas:
@piotrWichlinskiSoftwaremind can you rebase? It's bringing commits that were not supposed to be here.

src/browserCommands.ts
diocas: It's still not clear to me what the small changes are. Please annotate them.
Also provide the link to the file (in the specific commit) you took this code.

diocas:
Looks like this branch still has conflicts with master, so please rebase.

piotrWichlinskiSoftwaremind:
@diocas I've added more detailed comments in the code and rebased the branch

PR #117, opened 2022-03-23
optimize user search

dagl:
limit characters to min 3, when searching for a user, add parameter to skip searching user groups

This is a temporary fix, we are still working on doing the character limititation with validators. It's taking more time, because tornado doesn't implement them by default.

### DIFF
--- a/cs3api4lab/api/cs3_user_api.py
+++ b/cs3api4lab/api/cs3_user_api.py
@@ -45,7 +45,10 @@ class Cs3UserApi:
             return {}
 
     def find_users_by_query(self, query):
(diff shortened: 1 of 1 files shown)


diocas:
I guess it's ok to prevent listing for less than 3 chars in the backend, but why don't we do it from the frontend? This would give us the freedom to still do the full search if we wanted.

dagl:
@diocas Ok, so let's merge this

diocas (approved):

diocas:
Merging, to follow up the web part in #124

PR #127, opened 2022-04-20
Catching date parsing errors

mateusz-garlacz:

PR #128, opened 2022-04-27
Exposing root as homedir

mateusz-garlacz:
This fixes exposing slash for home directory. Redirecting to user directory will come next. This PR also contains #127 .

### DIFF
--- a/src/infobox.tsx
+++ b/src/infobox.tsx
@@ -211,17 +211,10 @@ const Shares = (props: ShareProps): JSX.Element => {
 
   useEffect(() => {
     const getGrantees = async (): Promise<any> => {
-      const home: any = await requestAPI<any>('/api/cs3/file/home', {
-        method: 'GET'
-      });
-
-      const resource =
-        home.path +
-        '/' +
-        props.fileInfo.path
-          .replace('cs3driveShareByMe:', '')
-          .replace('cs3drive:', '')
-          .replace('cs3driveShareWithMe:', '');
+      const resource = props.fileInfo.path
+        .replace('cs3driveShareByMe:', '')
+        .replace('cs3drive:', '')
+        .replace('cs3driveShareWithMe:', '');
 
       requestAPI<any>('/api/cs3/shares/file?file_path=' + resource, {
         method: 'GET'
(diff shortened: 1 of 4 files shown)


jupyter-config/jupyter_cs3_config.json
```diff
@@ -3,7 +3,7 @@
 	"reva_host": "127.0.0.1:19000",
 	"auth_token_validity": "3600",
 	"endpoint": "/",
-	"home_dir": "/home",
```
diocas, line 6: I wouldn't change the default, unless you think the shares will always need access to the `myshares` folder.. From other changes bellow, this seems to be what you're doing, is that correct?

src/infobox.tsx
```diff
@@ -211,17 +211,10 @@ const Shares = (props: ShareProps): JSX.Element => {
...
-      });
-
-      const resource =
-        home.path +
-        '/' +
-        props.fileInfo.path
-          .replace('cs3driveShareByMe:', '')
-          .replace('cs3drive:', '')
-          .replace('cs3driveShareWithMe:', '');
+      const resource = props.fileInfo.path
```
diocas, line 214: Why isn't this info removed from the backend instead? If I understand correctly, you already assume the drives are not exposed to the frontend, no?

diocas (commented):
Please rebase

jupyter-config/jupyter_cs3_config.json
mateusz-garlacz: done

src/infobox.tsx
mateusz-garlacz: After all there are no changes in this file. This is a matter of formatting, the replacing above is in master for months.

mateusz-garlacz: After all there are no changes in this file. This is a matter of formatting, the replacing above is in master for months.

dagl:
@diocas please take a look

src/infobox.tsx
diocas: I understand it was there, it was necessary, I just wonder with the changes done in the backend if this was address or not. But it's fine to keep it for a bit longer.

diocas: @mateusz-garlacz please let me know it it's necessary, otherwise let's merge.

mateusz-garlacz: @diocas it is still necessary to keep it

diocas (approved):

diocas:
@dagl @mateusz-garlacz was this tested? What config do I need to put? I've set `"home_dir": "/"` and when I try to navigate inside a folder, I get the message `Directory not found: "cs3drive:"` and the log message `[handlers:617] /eos is not a file`

diocas:
It was merged but I merged based on the changes seen, I didn't test it before. The problems are the ones I mentioned, I cannot navigate to folders from the root... Please check and open a PR with a fix.

dagl:
> 

@diocas We just checked and it works correctly for both me and @mateusz-garlacz on user ailldev. Maybe try building the plugin on a fresh virtual env?

PR #136, opened 2022-05-11
Apimanager refactor

dagl:
Remove duplicated code, refactor cs3apismanager

PR #139, opened 2022-05-22
Kernel path

dagl:
An optional setting for the get_kernel_path method in cs3apimanager

### DIFF
--- a/cs3api4lab/api/cs3_file_api.py
+++ b/cs3api4lab/api/cs3_file_api.py
@@ -42,6 +42,7 @@ class Cs3FileApi:
         intercept_channel = grpc.intercept_channel(channel, auth_interceptor)
         self.cs3_api = cs3gw_grpc.GatewayAPIStub(intercept_channel)
         self.storage_logic = StorageLogic(log)
+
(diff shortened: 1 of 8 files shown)


dagl:
Please megre after https://github.com/sciencemesh/cs3api4lab/pull/136

diocas (commented):
But in the meantime, I've already seen something that is no ok with what we discussed... We agreed that the kernel path should be dynamic, i.e allow username interpolation, so that I can do `/eos/user/{username[0]}/{username}`, for example.

But actually, I'm not sure if this is needed after all... (which means I can merge the PR as is)

I guess we can always do that dynamically on the config file itself and this is still acceptable (I'm thinking from a perspective of a person that installs the extension and configures it)...
We also need to remember that we can open a notebook that was shared with me... But I guess that with normal REVA the path will be built by appending "kernel_path" with the "myshares" directory, right? And for us, at CERN, I guess I we will have the `kernel_path=/`, because the files are already served in the UI from the full EOS path.

Do I miss something? Any comments?

diocas:
Discussed in the call, to be tested before merging

PR #140, opened 2022-05-24
hot-fix for move in cs3_file_api

dagl:


dagl:
@diocas currently moving files doesn't work, @gmgigi96 please run tests in the future before pushing code

diocas (approved):

PR #142, opened 2022-05-25
Fix kernel_path

diocas:
The behavior that was pushed was not correct, as it did not consider the directory of the notebook.

After some time debugging I also understood how this works... Jupyter takes the path where it starts and sets it as the "root_dir" variable  (if we do not specify it as a config). And then they use this in multiple places, including when setting the kernel path. Meaning that they combine `root_dir` with the output of `get_kernel_path` before changing the working dir of the kernel. 

So, when setting `kernel_path` we need to be extra careful... At least we can use relative paths (with e.g `../`).
We definitely need to document this.

I'm even wondering if we should have this config to begin with... We could explain that the Jupyter needs to be started in the local path that corresponds to what we expose as the root to the web and rely on the upstream `get_kernel_path` behavior (for now we need to remove the `drive:` that comes along, but @piotrWichlinskiSoftwaremind said this was going to be removed soon). What do you think?


dagl:
Your explanation makes sense, it's weird though, that returning the absolute path on windows worked. I was also considering this implementation from filecontentmanager, but if absolute path worked it made sense to me to just return it. 

Anyway I guess this should work, we also need to change the test but I can take care of that later

dagl:
@diocas setting relative path in the project directory doesn't seem to work for me, how exactly do you set the directory with slashes? And how do you test it?

dagl (changes requested):
@diocas setting relative path in the project directory doesn't seem to work for me, how exactly do you set the directory with slashes? And how do you test it?

dagl:
@diocas ok, I tested this once again and can confirm that it works

dagl (approved):
@diocas ok, I tested this once again and can confirm that this works

Issue #148, opened 2022-05-30
Use configurable traitlets


diocas:
Change the way we load the config and make it consistent with the Jupyter ecosystem.
I've asked this before, but we agreed to keep it as is for now. 

This would allow us to use some configs as callables instead of strings while self documenting all the options in a single place (thus fixing #109).

The ENV variables support can and should be kept.

Example: https://github.com/swan-cern/jupyter-extensions/blob/60e691b9b3ffe5241302fe909dd4f50ca8b7c040/SwanProjects/swanprojects/config.py

PR #165, opened 2022-05-30
Fixed running all tests at once

mateusz-garlacz:


cs3api4lab/config/config_manager.py
```diff
@@ -37,8 +37,16 @@ def __init__(self):
...
         self.config = self.config | config
 
-        # overwriting the values with env vars
-        env_vars = {env[4:].lower():os.environ[env] for env in os.environ if env.startswith("CS3_")}
+        # overwriting the values with env vars and parse boolean values
+        env_vars = {}
+        for env in os.environ:
+            if env.startswith("CS3_"):
+                if env[4:].lower() in ["secure_channel", "tus_enabled", "enable_ocm"]:
+                    value = os.environ[env] == "true"
```
diocas, line 45: Why don't you parse the string to a boolean? Otherwise, `CS3_SECURE_CHANNEL=false` will be set as true...

diocas: I'm not sure it wouldn't be better to just do #148 instead of this. It shouldn't take much time to do it and we just get it done.

cs3api4lab/tests/test_cs3_ocm_share_api.py
```diff
@@ -1,11 +1,12 @@
-from unittest import TestCase
+from unittest import TestCase, skip
 from cs3api4lab.api.cs3_file_api import Cs3FileApi
 from cs3api4lab.config.config_manager import Cs3ConfigManager
 from traitlets.config import LoggingConfigurable
 from cs3api4lab.api.cs3_ocm_share_api import Cs3OcmShareApi
 
 
-class TestCs3ShareApi(TestCase):
+@skip
```
diocas, line 8: I supposed we're skipping this just for the moment and we will bring this back soon, no?

mateusz-garlacz: yes

cs3api4lab/config/config_manager.py
mateusz-garlacz: I will take a look then

dagl: @diocas we will look at the https://github.com/sciencemesh/cs3api4lab/issues/148 a little bit later if that's ok, right now we want to focus on the higher priority tasks. https://github.com/sciencemesh/cs3api4lab/issues/148 is not as quick as it seems, there are a lot of places where config is used plus tests.

diocas: > @diocas we will look at the #148 a little bit later if that's ok, right now we want to focus on the higher priority tasks. #148 is not as quick as it seems, there are a lot of places where config is used plus tests.

Ok, no problem. I thought it was just a matter of replacing some things.

diocas: > Why is that? Any string other than 'true' should be evaluated as false

Ah sorry!! I misread the code!!

diocas (approved):

PR #166, opened 2022-05-31
fix loading shares without active shares tab

### LINKED ISSUE #161: Loading shares when the shares panel is not opened
I see the call to the shares API (including for update) being done even if I'm not in the shares panel. This should not happen, especially since this is extremely expensive and even blocks the main thread.

piotrWichlinskiSoftwaremind:
fix loading pending shares panel after a tab is active, fix decline icon, fix pending share element unique key error

fixes #161


piotrWichlinskiSoftwaremind:
This PR should solve loading share issues in issue #161

PR #167, opened 2022-05-31
Setting user in user directory

mateusz-garlacz:


cs3api4lab/api/cs3apismanager.py
```diff
@@ -8,10 +8,17 @@
...
 from tornado import web
 from nbformat.v4 import new_notebook
-from cs3api4lab.config.config_manager import Cs3ConfigManager
 from cs3api4lab.utils.share_utils import ShareUtils
 from cs3api4lab.api.share_api_facade import ShareAPIFacade
 from cs3api4lab.utils.model_utils import ModelUtils
+import os
+import json
+
+from cs3api4lab.utils.file_utils import *
```
diocas, line 17: Like LGTM, I also don't like this

cs3api4lab/api/cs3apismanager.py
```diff
@@ -25,6 +32,18 @@ def __init__(self, parent, log):
         self.log = log
         self.file_api = Cs3FileApi(self.log)
         self.share_api = ShareAPIFacade(log)
+        self._init_user_dir_setting()
+
+    def _init_user_dir_setting(self):
+        if "user_dir" in self.cs3_config and self.cs3_config["user_dir"] != "":
```
diocas, line 38: Uhm... Not sure if I would document this instead of saving this config file.
Also, since we have a server extension that exposes an API, why don't we just create a new endpoint to return the session info with the user_dir there?

README.md
```diff
@@ -186,6 +186,8 @@ Config file fields:
 - client_id - client login to authenticate in Reva
 - client_secret - client password to authenticate in Reva
 - root_dir_list - list of root dirs, for example https://developer.sciencemesh.io/docs/iop/deployment/kubernetes/providers/ root dirs are "/home,/reva"
+- home_dir - home directory of the user
```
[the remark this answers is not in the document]
diocas: Maybe make one the "mount_dir" and the other "home_path"?

cs3api4lab/api/cs3apismanager.py
diocas: Then we don't need to care about the config system, which we know can be modified by the user.

src/index.ts
```diff
@@ -397,6 +397,19 @@ const cs3browser: JupyterFrontEndPlugin<void> = {
         fileBrowser.useFuzzyFilter = useFuzzyFilter;
       });
 
+    /**
+     * cs3api4lab modification - loads plugin settings to redirect user to user directory at start
+     */
+    fileBrowser.model.restored.then(() => {
```
diocas, line 403: I would even add a "home" button in the toolbar for users to get to the starting page more easily.

diocas (changes requested):

src/index.ts
mateusz-garlacz: You are correct, first / is requested and then it is switched to user dir. 
I would add another button in separate task though.

README.md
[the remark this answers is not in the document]
mateusz-garlacz: how about "mount_dir" and "home_dir"?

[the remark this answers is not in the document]
diocas: sounds good!

src/index.ts
diocas: Yes, done #168 

For the cd, can't we change the first directory instead of doing cd? would that mean big changes or not really?

piotrWichlinskiSoftwaremind: @diocas A button in the toolbar for redirecting to user home dir is even better solution than automatic redirection. Especially when file browser automatically changes the directory to an opened document and remembers the path of the directory when the page is refreshed. So if you button instead of automatic redir it's fine with me.

cs3api4lab/utils/file_utils.py
```diff
@@ -37,3 +37,14 @@ def calculate_content_size(content):
             content_len = len(content.decode('utf-8'))
         content_size = str(content_len)
         return content_size
+
+    @staticmethod
+    def get_user_directory(config):
```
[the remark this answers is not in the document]
piotrWichlinskiSoftwaremind: @diocas  If you can generate a path to the user directory and pass it into the plugin config we can skip this part with generating the path by ourselves. user directory should still be optional or contain the same value as home_dir in case of other environments especially our local dev environments

cs3api4lab/api/cs3apismanager.py
piotrWichlinskiSoftwaremind: @diocas This solution was based on our talk about the standard jupyter way of passing settings to the frontend, but of course we can rework it to use a simple endpoint returning the user home directory

diocas: Yes, we discussed this. But I though, from what I understood of your explanation, that this was somehow being used by the upstream code to change the user home dir.
In the end, you added some code that `cd`'s into the directory. So, if that is the case, I think this approach brings more complexity without any advantage compared to just calling an API endpoint.

cs3api4lab/utils/file_utils.py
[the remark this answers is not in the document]
diocas: Yes, I can generate the path of the user and put it in the config. So please skip this part and return it as is.
My only issue is that this is not properly configurable, it's a hardcoded path in the end.

src/index.ts
diocas: The button is complementary, we should always start in home if there is no other path loaded. But this is a good point, with this cd we're not even taking that in consideration, right? Can't you see how they do it and do the same?
I would expect the ui to open in some path if it's given, or in home (or `/`) is not.

piotrWichlinskiSoftwaremind: @diocas I tested this on the dev environment and it is taking other redirection into consideration when the document is opened or the user changed directories and refreshed the page our redirection will be skipped and it will switch to the current directory. Only thing that I'm worried about is loading times, I can take a look if there is a way to control this better but the default behavior is coded into file browser plugin and I'm not sure we can fully control it. I will take a look what else we can do with that.

diocas (approved):

PR #171, opened 2022-06-08
Drives names removal

### LINKED ISSUE #156: Use relative paths
Remove "cs3drive:" from the UI and from everywhere and just use relative paths. Any modification to the path should be done in the backend and only there.

Needs to be double checked in relation to the OCM shares (####), but I'm assuming now that those will be mounted as the other shares. To be seen.

mateusz-garlacz:
Closes #156


dagl:
https://github.com/sciencemesh/cs3api4lab/issues/156

diocas (approved):

PR #172, opened 2022-06-13
add python tests workflow, cleanup tests 

dagl:


dagl:
@diocas things to consider:

- caching venv instead of pip cache is about 30s faster per run, but maybe there are some downsides?
- for some reason only the latest revad docker image works, any other tag I tried ends up with errors
- also sometimes Reva doesn't create it's own configs correctly and only works after being restated - hence the 'check Reva' step
- revad can be run directly from docker run instead of docker compose if you prefer

diocas:
"Solves" is not a valid closure tag. I've commented on the traitlets PR.

diocas:
The way this task was done doesn't seem at all with what is described in the link I attached in the ticket. And it seems a much worse solution.
Did you try it and was not working, or is there any reason to do it like this?

dagl:
@diocas I looked at this but I don't see any way to make it possible with the default revad image, because of the configs we use ([examples/two-server-setup](https://github.com/cs3org/reva/tree/master/examples/two-server-setup)), which this image doesn't have. It would require a custom dockerfile and image.

Issue #109, opened 2022-02-14
Different lists of config variables in a few places


michzimny:
README lists dozen of environmental variables near its end. README also refers to [reva_config.env](../blob/master/reva_config.env) that lists some variables. 

However, the lists of variables in those two places seems to be different. Moreover, they seem to be different than in [config_manager.py](../blob/master/cs3api4lab/config/config_manager.py).

It should be unified.

PR #173, opened 2022-06-13
Traitlets config

mateusz-garlacz:


dagl:
https://github.com/sciencemesh/cs3api4lab/issues/148

mateusz-garlacz:
#109

diocas:
To close issues automatically, please add "Fix #xxx" or "Closes #xxx" in the PR description directly (or on a comment). More info [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

diocas:
As for the PR, instead of documenting this in the README, I would move the documentation directly to the config file itself. And I would set the traitlets properly, as described in the ticket. At the moment this is not the case. If you want, I can try to merge this in another branch and have a look at it, but as is not ok.

dagl:
@diocas Thanks for the info on linking in GH. As for the traitlets, indeed now I see it wasn't done correctly, we'll fix that.

diocas:
@dagl basically, just have a class with those variables. Put the description directly on them and and set types so that we don't need to check them or do conversions. Then we're sure we're using the proper types.
For the env vars, put them as the "default" values. You can use `os.getenv` as a default, which means we will 1) check the config, then the ENV and lastly a default values (set in getenv). Unless you think it's better to make the ENV the first to be checked, then I need to think how to do it.

dagl:
@diocas Ok thanks. Today is public holiday in Poland and tomorrow Mateusz is on vacation so we'll take a look on Monday.

dagl:
> @dagl basically, just have a class with those variables. Put the description directly on them and and set types so that we don't need to check them or do conversions. Then we're sure we're using the proper types. For the env vars, put them as the "default" values. You can use `os.getenv` as a default, which means we will 1) check the config, then 2) the ENV and lastly 3) a default value (set in getenv). Maybe it's better to make the ENV the first to be checked, then I need to think how to do it.

Hmm if we did it like this (default/env as the last) would teh default value override config if env is not set? I think we should set default as a hardcoded value, and first set the default value, then config (if it exists) and then env (if exists). This way we ensure values are set in the right order.

mateusz-garlacz:
Hi, @diocas - ready for CR

cs3api4lab/config/config_manager.py
```diff
@@ -2,14 +2,216 @@
 from jupyter_core.paths import jupyter_config_path
 from notebook.services.config import ConfigManager
 from traitlets.config import LoggingConfigurable
+from traitlets import Unicode, Bool, CInt, Tuple, default
 
 
 class Config(LoggingConfigurable):
-    config = {
-        "auth_token_validity": "3600",
+
```
mateusz-garlacz, line 9: please let me know if this display i.e. the list of variables before constructors is suitable

diocas (approved):

PR #175, opened 2022-06-23
Traitlets config

### LINKED ISSUE #109: Different lists of config variables in a few places

### LINKED ISSUE #148: Use configurable traitlets

diocas:
Replacement for #173: I've rebased on the latest master after merging the drives names PR. So it should be exactly the same (with addition of `mount_dir`).

Once #174 gets merged, I'll rebase and see if the tests succeed (but feel free to have a look if you have time).

Closes #148 
Closes #109


mateusz-garlacz (approved):

mateusz-garlacz (changes requested):
Hi, something went wrong with this rebase, it does not build. I rebased my branch SoftwareMind-ScienceMesh-org:traitlets_config and resolved conflicts. Please let me know if I can recreate this https://github.com/sciencemesh/cs3api4lab/pull/173

diocas:
@mateusz-garlacz yep, feel free to re-open the other and I'll close this

diocas:
But we still need to rebase on top of the tests, which is not merged yet.

PR #176, opened 2022-06-28
is_dir optimisation

mateusz-garlacz:
cs3apismanager.is_dir method optimisation and connected changes

### DIFF
--- a/cs3api4lab/api/cs3_file_api.py
+++ b/cs3api4lab/api/cs3_file_api.py
@@ -17,7 +17,7 @@ import cs3.storage.provider.v1beta1.provider_api_pb2 as cs3sp
 from cs3api4lab.exception.exceptions import ResourceNotFoundError
 
 from cs3api4lab.utils.file_utils import FileUtils
-from cs3api4lab.logic.storage_logic import StorageLogic
+from cs3api4lab.api.storage_api import StorageApi
 from cs3api4lab.auth import check_auth_interceptor
 from cs3api4lab.auth.authenticator import Auth
 from cs3api4lab.auth.channel_connector import ChannelConnector
@@ -41,7 +41,7 @@ class Cs3FileApi:
         auth_interceptor = check_auth_interceptor.CheckAuthInterceptor(log, self.auth)
         intercept_channel = grpc.intercept_channel(channel, auth_interceptor)
         self.cs3_api = cs3gw_grpc.GatewayAPIStub(intercept_channel)
-        self.storage_logic = StorageLogic(log)
+        self.storage_api = StorageApi(log)
 
         self.lock_manager = LockManager(log)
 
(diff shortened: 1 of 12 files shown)


diocas:
@mateusz-garlacz can you rebase?

diocas:
@mateusz-garlacz after the merger of the traitlets, it's again necessary to rebase... I also still need to review.

cs3api4lab/api/cs3_file_api.py
```diff
@@ -57,53 +57,52 @@ def mount_point(self):
...
         Stat a file and returns (size, mtime) as well as other extended info using the given userid as access token.
         Note that endpoint here means the storage id. Note that fileid can be either a path (which MUST begin with /)
         or an id (which MUST NOT start with a /).
         """
         time_start = time.time()
-        ref = FileUtils.get_reference(file_id, endpoint)
-        stat_info = self.cs3_api.Stat(request=cs3sp.StatRequest(ref=ref),
-                                      metadata=[('x-access-token', self.auth.authenticate())])
+        
         time_end = time.time()
```
diocas, line 68: You don't do anything in between start and end. Doesn't seem right...

cs3api4lab/api/cs3_file_api.py
```diff
@@ -57,53 +57,52 @@ def mount_point(self):
...
     def read_file(self, file_path, endpoint=None):
         """
         Read a file using the given userid as access token.
         """
         time_start = time.time()
-        
-        if self.storage_logic.stat(file_path, endpoint) is not None:
+
+        stat = self.storage_api.stat(file_path, endpoint)
+        if stat.status.code == cs3code.CODE_OK:
```
diocas, line 98: Why don't we bail if code not ok? Otherwise, you will try to download a file whose stat failed.

cs3api4lab/api/cs3_file_api.py
```diff
@@ -129,14 +128,15 @@ def write_file(self, file_path, content, endpoint=None):
 
         time_start = time.time()
 
-        if self.storage_logic.stat(file_path, endpoint):
+        stat = self.storage_api.stat(file_path, endpoint)
+        if stat.status.code == cs3code.CODE_OK:
```
diocas, line 132: Here it makes sense to write even is stat fails, but only for file not existing reasons. If the API fails, maybe we should fail as well.

cs3api4lab/api/cs3_file_api.py
```diff
@@ -210,7 +210,8 @@ def move(self, source_path, destination_path, endpoint=None):
         src_reference = FileUtils.get_reference(source_path, endpoint)
         dest_reference = FileUtils.get_reference(destination_path, endpoint)
 
-        if self.storage_logic.stat(destination_path, endpoint):
+        stat = self.storage_api.stat(destination_path, endpoint)
+        if stat.status.code == cs3code.CODE_OK:
```
diocas, line 214: Can it be that is not the file that exists, but the call failed?

diocas (changes requested):

cs3api4lab/api/cs3_file_api.py
dagl: Right now Reva sometimes returns code internal for non-existing files

cs3api4lab/api/cs3_file_api.py
dagl: not sure what you mean here, if the call fails then subsequent calls will also probably fail

diocas: I was wondering if it's possible that the call failed for some reason but the file actually exists. The next call will do a move on top of a file.
But, as you say, if it's not clear from the error code wether the file does not exist or smth else failed, I guess we should continue...

cs3api4lab/api/cs3_file_api.py
diocas: I was even going to ask why we even stat here and not just try to lock. But I guess that when we switch to the lock api, we need the file to exist, since the lock is written in its metadata.

If reva is not consistent, I guess there's no other alternative and I will resolve this discussion, but please add a FIXME here and explain the situation.. Cause we can be overwriting a file just because some network request failed.

cs3api4lab/api/cs3_file_api.py
diocas: This is the same question as above

cs3api4lab/api/lock_manager.py
```diff
@@ -74,10 +75,10 @@ def resolve_file_path(self, file_path, endpoint):
...
 
     def _resolve_directory(self, dir_path, endpoint):#right now its possible to write in somone else's directory without it being shared
-        try:
-            self.storage_logic.stat(dir_path, endpoint)
+        stat = self.storage_api.stat(dir_path, endpoint)
+        if stat.status.code == cs3code.CODE_OK:
             return dir_path
-        except Exception:
+        else:
             return self.config.mount_dir + '/'
```
diocas, line 82: This might lead to conflicts between files with the same name. I was going to ask if we can de-duplicate, but since we will switch to the lock api, doesn't seem worth the trouble.

cs3api4lab/api/cs3_file_api.py
mateusz-garlacz: added fixme in both places

mateusz-garlacz:
@diocas rebased

diocas (approved):

diocas (approved):

PR #179, opened 2022-07-20
Removed notebook imports

### LINKED ISSUE #63: Remove Notebook imports
After upgrading to Jupyter 3 the plugin is compatible with jupyter-server and we should abandon notebook dependencies.

mateusz-garlacz:
Closes #63


diocas:
@mateusz-garlacz this comes at the perfect time... I was running the tests and they were failing and I suspect Jupyter has released a new version without the dependency on notebook and the installation we're doing in the CI didn't have it. I'll merge this and then the tests.

diocas (approved):

Issue #159, opened 2022-05-30
Listing incoming shares fails


diocas:
Listing received shares sometimes works, others doesn't. Needs to be understood and a proper log message should be given.

mateusz-garlacz:
Hi, so instead of showing the directory, maybe we could list received shares with accepted status. Just like the 'pending' panel shows shares with pending status. Basically we would get back to the previous solution with invoking stat for every item, but I could add some simple caching with configurable TTL, like e.g.:
https://github.com/sciencemesh/cs3api4lab/blob/baa69f0635d4336bf11011f4afd7a9a9d82a7bfc/cs3api4lab/api/lock_manager.py#L103
The preffered solution (automatic shares acceptance) would require frontend adjustments and some backend work (for which there is no time estimation I guess?), so I assume this would make listing incoming shares work as a in-between solution.

@diocas

PR #182, opened 2022-07-29
Hiding breadcrumbs

mateusz-garlacz:


cs3api4lab/utils/file_utils.py
```diff
@@ -46,10 +46,58 @@ def normalize_path(path):
...
 
+    @staticmethod
+    def get_my_shares_path_prefix(config):
+        if config.mount_dir == "/":
+            return config.home_dir + "/" + config.shared_folder
+        else:
+            return "/" + config.shared_folder
+
+    @staticmethod
+    def get_share_path_prefix(config):
```
diocas, line 57: @gmgigi96 can you check this and the rest? Is it ok to assume shares will be in `/reva/etc`?
Also, how is this handle drives in revacmd? Because it only returns paths without any drives, AFAIR, so do we also do these replaces there?

cs3api4lab/utils/file_utils.py
```diff
@@ -46,10 +46,58 @@ def normalize_path(path):
...
+                drive = "/" + drive
+            if path.startswith(drive):
+                return drive
+        if path.startswith("/cs3drive") or path.startswith("cs3drive"):
+            return "/cs3drive" if has_root else "cs3drive"
+        return None
+
+    @staticmethod
+    def replace_drive_with_directory(path, drive, config):
+        path = path.replace(":", "")
```
diocas, line 92: Why don't we replace bellow when replacing the drive? (can we have a path with ":"? if so, we would be replacing too much)

cs3api4lab/api/share_api_facade.py
```diff
@@ -128,6 +130,9 @@ def list_grantees_for_file(self, file_path):
         :param file_path: path to the file
         :return: list of grantees
         """
+        file_path = FileUtils.handle_drive(file_path, self.config)[0]
+        file_path = file_path.replace(":", "")
```
diocas, line 134: isn't this replaced done inside handle_drive already?

cs3api4lab/utils/file_utils.py
```diff
@@ -46,10 +46,58 @@ def normalize_path(path):
...
+        if "cs3driveShareWithMe" in drive:
+            prefix = FileUtils.get_my_shares_path_prefix(config)
```
diocas, line 94: Not sure why we don't use `get_my_shares_path_prefix` (instead of duplicating code) and do a single `path = path.replace(drive, prefix)`. The logic seems the same for all 3 cases, which is you always replace drive with the prefix.

setup.py
```diff
@@ -59,7 +59,8 @@
         "grpcio",
         "grpcio-tools",
         "PyJWT==2.4.0",
-        "webdavclient3"
+        "webdavclient3",
+        "notebook"
```
diocas, line 63: We should not depend on notebook anymore...

src/drive.ts
```diff
@@ -74,7 +74,13 @@ export class CS3Contents implements Contents.IDrive {
     const activeTab: string = (await this._state.fetch('activeTab')) as string;
```
diocas, line 75: We agreed that we would not use drive names in the frontend anymore, so why are we adding it again? Do we really need it? Wouldn't we be able to detect from the path instead of having to append drives and then check them in the backend to do a replace?

diocas (changes requested):

cs3api4lab/utils/file_utils.py
mateusz-garlacz: On cernbox the shares path are "/eos/user/x/...", on local reva setup the shares paths are "/reva/user/...", I am unable to make it more generic

src/drive.ts
mateusz-garlacz: Yes, unfortunately we need it to make it work. The reason is in order to hide breadcrumbs I have to send incomplete file path in dir model to frontend, when the user clicks on a file, the request going to backend contains that incomplete path and without the name of the drive included it is impossible to determine what prefix should be added

cs3api4lab/utils/file_utils.py
mateusz-garlacz: Colons are prohibited in file names. Colons are sometimes glued automatically to the drive.

diocas: But the question remains, why don't re replace bellow?

mateusz-garlacz:
After some consideration, I think we should resolve #159 first, because if we go back to listing backend response instead of directory, this PR needs to change

cs3api4lab/utils/file_utils.py
```diff
@@ -46,10 +46,58 @@ def normalize_path(path):
             path = '/'
         return path
 
+    @staticmethod
+    def get_my_shares_path_prefix(config):
```
diocas, line 50: This was discussed by email and it's not correct that the shares folder lives inside the home dir. We also discussed that we cannot use this folder any longer, so I believe this PR needs to be changed.

cs3api4lab/utils/file_utils.py
diocas: This comment still stands

diocas (changes requested):

mateusz-garlacz:
I am closing this PR due to incompatibility - after abandoning the shared folder approach there is no way to determine and recreate the full path on backend with the requested shared file name

PR #185, opened 2022-08-23
share form changes

piotrWichlinskiSoftwaremind:
fixed all tests for shares and ocm
fixed css for share form,
reconfigure select component for search users/permissions, 
fix remove share/ocm (backend),
add remove share/ocm (frontend)
fix get grantees, add ocm grantees, 
fix remove share/ocm, 
add filters by resource to share/ocm list,
add the username to the response (needed for the select box)
fix create ocm - incorrect resource id
fix check if the share is public or ocm
separate get ocm from the list
Share is created automatically when selecting a user with viewer permissions
add change permissions on frontend
added username for shares select users/grantees list
add find user by username/email 
filter repeating users in the list (front)
use react hooks to add debouncer
fix bug with getting the correct value in the search box
bump cs3apis to 0.1.dev94


diocas (commented):
I left some comments pending, but I think they are still valid. It's missing documentation that explains what methods accept and what some do.

cs3api4lab/api/cs3_ocm_share_api.py
```diff
@@ -147,21 +144,28 @@ def update_received(self, ocm_share_id, field, value):
             self._handle_error(response, "Error updating OCM received share:")
         return response.opaque
 
-    def list(self, share_id=None):
-        if share_id is None:
-            return self._list_ocm_shares()
-        else:
-            return self._get_ocm_share(share_id)
+    def list(self, optional_filter={}):
```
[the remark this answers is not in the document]
piotrWichlinskiSoftwaremind: > I guess this way is to retrieve just a specific resource instead of a list with all of them?

Yes, this is exactly what it does. I have noticed that in a lot of places in our plugin we are not using filters implemented by Reva API. Most of the time we are getting a whole shared file list (even when there is a need only for information on one file) for a user and then filtering them by a path, that design is heavy on load and is not "bulletproof". By using filters we can cut down the load and receive from Reva only shares that we need, especially that shared files are identified by resource id (storage_id, opaque_id). This also resolves some problems like how to determine share and OCM, previous implementation was bugged.

cs3api4lab/api/share_api_facade.py
```diff
@@ -128,23 +132,30 @@ def list_grantees_for_file(self, file_path):
         :param file_path: path to the file
         :return: list of grantees
         """
-        share_list = self.share_api.list()
+
+        file_path = FileUtils.remove_drives_names(file_path)
+        file_path = FileUtils.check_and_transform_file_path(file_path)
+        file_stat = self.file_api.stat_info(file_path)
+        share_filter = {
```
diocas, line 139: This is what I was talking with you, I don't like the abstraction this high in the stack. These CS3APIs details should be hidden as much as possible lower, like in the cs3 client, and then we can expose something a bit more high level.
Why don't we just call a  new method, i.e `self.share_api.file_shares`, and pass the file path (or the stat info), and we build the filter inside?

cs3api4lab/api/cs3_ocm_share_api.py
diocas, line 147: Then in this method, the optional filter parameter is passed directly to `ListOCMSharesRequest`.

Otherwise, you are building a method that accepts some kind of filtering, which is great, but in reality you only accept a resource filter (which needs to be converted).

So, this method is generic, and you create a separate method (the `self.share_api.file_shares` I mentioned above) that calls this one and sets the filter appropriate to the use case.

diocas (changes requested):
I think this is a quick change, is just moving the code to a new method. Let me know if I'm not being clear.

diocas (approved):

PR #190, opened 2022-09-04
Async requests

dagl:
First of all, I didn't fully implement async grpc calls, and instead opted for async wrappers for (almost) all of our plugin's handlers. Full port to async can be done - by changing all usages of grpc._channel to grpc.aio._channel - however it's not straight forward and would have taken much more time. Some of the methods from the grpc module that we use for intercepting grpc calls and authentication aren't present in the grpc.aio library, so we would have to write our own implementations. 

Second, we need to keep in mind that everything in our plugin runs from a synchronous context.

The solution I came up with is simple - first we make all of our synchronous server methods run in async, so that they don't block each other, and then we have to somehow run those async methods from the sync context, in a way that doesn't block the main thread. 

I did this in two different ways. In handlers.py we have full control of all of our own methods, so we can do this natively (https://www.tornadoweb.org/en/stable/guide/async.html):

- First we make a normal synchronous method run as an async method:

  `           loop = get_or_create_eventloop()`
`              result = await loop.run_in_executor(None, some_sync_method, *args)
  `
            
- Then we run this new async method from the sync context by using yield:

  `           yield RequestHandler.async_handle_request(self, self.user_api.find_users_by_query, 200, query)`

But when it comes to /api/contents methods, things get a little bit more complicated. We don't have direct access to the tornado handlers, so instead I created the asyncify wrapper - it basically does the same thing but uses `loop.run_until_complete(run_async(loop, *args, **kwargs))` instead of yield, to run async from sync context, as yield is a proprietary way of doing async in tornado. 

There is a small caveat - every time we modify or create a notebook, the in-memory SQLite database gets updated and it doesn't allow operations from different threads. So all of notebook methods in cs3apismanager.py remain synchronous for now. We could make SQLite accept updates from different threads by changing check_same_thread flag to false when the connection to DB is established: `sqlite3.connect(db_file, check_same_thread=False)`, This would require copying the whole nbformat/sign.py implementation, changing the flag here: https://github.com/jupyter/nbformat/blob/main/nbformat/sign.py#L147 and then pointing jupyter to our new implementation in the jupyter_server_config.py, the same way we point to our cs3apismanager implementation. I tested changing the flag in the downloaded library and it seems to work, so let me know if you'd like to try this.

There is a lot going on so if you would like to have a call about this PR let me know on gitter.


cs3api4lab/api/cs3apismanager.py
```diff
@@ -15,7 +15,7 @@
 from cs3api4lab.utils.file_utils import FileUtils
 from cs3api4lab.api.share_api_facade import ShareAPIFacade
 from cs3api4lab.utils.model_utils import ModelUtils
-
+from cs3api4lab.utils.asyncify import asyncify
 
 class CS3APIsManager(ContentsManager):
```
diocas, line 20: When I reported the issue, I didn't necessarily see any issues with these requests (contents manager), if I'm not mistaken, but with the calls to our new handlers (for the cs3 apis). So I'm not sure this is necessary here, but we can leave it to try.

diocas (approved):
Your assumptions seem fair and I don't see a problem with the solution. The problem was the blockage of the main thread, so I don't think we necessarily need the calls themselves to be async, so I think this is ok as it is.

As I mention in a comment bellow, I don't thinkg we had the need to make the contents manager calls async, but let me test as is.

cs3api4lab/api/cs3apismanager.py
dagl: It helps with reading directories, opening files etc

PR #191, opened 2022-09-04
Preventing OCM call failure

mateusz-garlacz:


cs3api4lab/api/share_api_facade.py
```diff
@@ -162,11 +162,16 @@ def is_ocm_share(self, share_id):
 
     def is_ocm_received_share(self, share_id):
         """Checks if share is present on OCM received shares list"""
-        received_shares = self.ocm_share_api.list_received()
-        for share in received_shares.shares:
-            if share_id == share.share.id.opaque_id:
-                return True
-        return False
+        try:
```
diocas, line 165: Shouldn't we put the check of the config `enable_ocm` instead of this (or with this as well)?

I only see this method being called in one place. We actually check that config there, but only if this method returns true. I guess we should check it before to make it consistent.

This is should be ok because I don't think we will have OCM shares and this config set to false, otherwise I hope the attempt to update a normal doesn't break anything.

diocas (changes requested):

cs3api4lab/api/share_api_facade.py
mateusz-garlacz: @diocas I made a change and rebased

diocas (approved):

PR #198, opened 2022-09-20
fix opening base64 files

### LINKED ISSUE #195: Cannot open non text files
It fails in the backend decoding utf8, which shouldn't be trying to do in the first place.
Test with a pdf or picture, for example.

piotrWichlinskiSoftwaremind:
Fixes #195

### DIFF
--- a/cs3api4lab/api/cs3_file_api.py
+++ b/cs3api4lab/api/cs3_file_api.py
@@ -5,7 +5,6 @@ CS3 File API for the JupyterLab Extension
 
 Authors:
 """
-
 import http
 import time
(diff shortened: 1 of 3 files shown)


diocas (approved):

diocas (changes requested):
I've approved, but then I saw the upstream implementation is not exactly the same: https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/contents/fileio.py#L301-L328

They basically invert the if, and I guess their's might be less error prone.

diocas (approved):

PR #209, opened 2022-11-25
Added grpc locks support

mateusz-garlacz:


cs3api4lab/api/cs3_file_api.py
```diff
@@ -125,22 +127,20 @@ def write_file(self, file_path, content, endpoint=None, format=None):
         Write a file using the given userid as access token. The entire content is written
         and any pre-existing file is deleted (or moved to the previous version if supported).
         """
-        file_path = self.lock_manager.resolve_file_path(file_path, endpoint)
-
+        # todo fix me
+        #file_path = self.lock_api.resolve_file_path(file_path, endpoint)
```
diocas, line 131: Does this mean that _for now_ we're not creating conflict files?

cs3api4lab/exception/exceptions.py
```diff
@@ -1,77 +1,87 @@
...
-        self.message = "Missing argument: " + str(key_error)
-        super().__init__(self.message)
-
-    def __str__(self):
+class ShareError(Exception):
+    def __init__(self, message):
+        self.message = message
+        super().__init__(self.message)
+
+    def __str__(self):
```
diocas, line 6: I'm not sure why this appears as changed, but can you create a new PR where you fix these exceptions?

Instead of repeating the `__str__` on all of them, just create a new class, that inherits from `Exception`, and use that one as the parent of all the others.

cs3api4lab/locks/cs3.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def is_valid_external_lock(self, file_path, endpoint):
+        ref = FileUtils.get_reference(file_path, endpoint)
+        lock = self._get_lock(ref)
+        if lock:
+            return self._is_lock_expired(lock) and not self._is_lock_mine(lock)
+        return False
+
+    @staticmethod
+    def _is_lock_expired(lock):
```
glpatcern, line 45: This method seems to be both not needed (the storage layer never returns an expired lock) and incorrect, as the lock internal properties are not accessible like this. You could compare with the wopiserver implementation at https://github.com/cs3org/wopiserver/blob/master/src/core/cs3iface.py#L215-L226

cs3api4lab/locks/metadata.py
```diff
@@ -0,0 +1,69 @@
...
+        return lock and not is_mine and not self._is_lock_expired(lock)
+
+    def _is_lock_mine(self, lock):
+        user = self.get_current_user()
+        if lock:
+            return lock['username'] == user.username and lock['idp'] == user.id.idp and lock[
+                'opaque_id'] == user.id.opaque_id
+        return False
+
+    def _is_lock_expired(self, lock):
```
glpatcern, line 59: see the above comment for how to refer to the properties of a lock

glpatcern (commented):
I had a look at this code given the similarity with the wopiserver logic for locks, but as I don't know all the code, please take the following comments with a grain of salt.

Yet, despite the unit tests, the lock payload seems not to be handled correctly according to the CS3 APIs.

cs3api4lab/locks/cs3.py
```diff
@@ -0,0 +1,95 @@
...
+            self._unlock(ref, lock)
+            self._set_lock(ref)
+        else:
+            raise FileLockedError("File %s is locked" % file_path)
+
+    def is_valid_external_lock(self, file_path, endpoint):
+        ref = FileUtils.get_reference(file_path, endpoint)
+        lock = self._get_lock(ref)
+        if lock:
+            return self._is_lock_expired(lock) and not self._is_lock_mine(lock)
```
diocas, line 41: Why is an expired lock a valid external lock? I'm confused.

If the lock is not mind and has NOT expired, then I want to create a conflict, right?

diocas: This even goes against the logic for the metadata implementation:

```python
lock and not is_mine and not self._is_lock_expired(lock)
```

cs3api4lab/locks/cs3.py
```diff
@@ -0,0 +1,95 @@
...
+class Cs3(LockBase):
+
+    def __init__(self, log, config):
+        super().__init__(log, config)
+        self.user_api = Cs3UserApi(log)
+        self.config = config
+
+    def handle_locks(self, file_path, endpoint):
+        ref = FileUtils.get_reference(file_path, endpoint)
+        lock = self._get_lock(ref)
```
diocas, line 26: I see that the naming of the internal methods are not consistent (i.e `_set_lock` vs `_lock_file`).
They don't necessarily need to be, but I would prefer if they were. It makes things easier to understand. Please check as well that we can make all methods aligned.
(Ofc this is not urgent, but I don't expect it to be a big change...)

diocas (commented):
I also asked @glpatcern to have a look. But seems fine, with exception of that if...

cs3api4lab/exception/exceptions.py
diocas: For later, ofc

cs3api4lab/api/cs3_file_api.py
mateusz-garlacz: Correct, creating conflict versions will be addressed separately, currently they are flawed and need some work

cs3api4lab/exception/exceptions.py
mateusz-garlacz: noted

cs3api4lab/locks/metadata.py
```diff
@@ -0,0 +1,69 @@
...
+                return
+            if self._is_lock_expired(lock):
+                self._lock_file(file_path, endpoint)
+                return
+        raise FileLockedError("File %s is locked" % file_path)
+
+    def _generate_lock_entry(self):
+        user = self.get_current_user()
+        return {self.lock_name: urllib.parse.quote(json.dumps({
+            "username": user.username,
```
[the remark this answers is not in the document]
mateusz-garlacz: Hi! This is our custom implementation of locks which uses file's arbitrary metadata entries. The CS3 implementation present in cs3.py does not include user's info manually.

cs3api4lab/locks/cs3.py
mateusz-garlacz: good point

cs3api4lab/locks/cs3.py
mateusz-garlacz: The lock is mapped to json first, so the props are accessible this way. I will take a look if it's needed anyway

cs3api4lab/locks/metadata.py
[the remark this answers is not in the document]
glpatcern: OK, but then how is the `Cs3._set_lock` method going to get the `lock_name` content? I'm probably missing something in the inheritance, but it seems the `Cs3` class implementation relies on this inherited function to actually create the payload.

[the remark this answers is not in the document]
mateusz-garlacz: _lock_name_ is defined in the base class, and this will be used as _lock_id_ in case of cs3 implementation, in case of metadata implementation we need to manually include user's info into the entry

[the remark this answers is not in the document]
glpatcern: Right, so my original comment holds: the `lock_id` is going to include details that are redundant compared to the metadata implementation. And also, when returning the lock in https://github.com/sciencemesh/cs3api4lab/pull/209/files#diff-828df65314d9ed749fee75882a0cc9a64a2bea77bbc9493a4b37a9c667b2f9bfR44 as the whole lock is returned, not just `lock_id`, the dict you obtain is not equivalent to the content you'd get from the pure metadata implementation.

My suggestion here would be to actually NOT try to be compatible and maintain both implementation, but rather adopt the native locks as they are and simplify the whole logic (no need to check for expired locks, etc.). Or if really you want to keep the two implementation, then `Cs3._get_lock` should return `lock.lock_id`, but the code looks a bit cluttered with redundant logic.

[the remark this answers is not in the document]
mateusz-garlacz: I already corrected checking expiration time in _Cs3_ unnecessarily, do you see another overcomplications in _Cs3_ implementation?

And maybe I didn't highlighted it, but the methods like _set/_getlock etc though named the same, are not a part of the interface defined in base class, and in each case the locks are generated differently, though they share a common lock_id. So in case of _Metadata_ leaving only _created_ and _updated_ will make impossible to determine the lock owner. 

Wether we need to maintain 2 implementations it's up to @diocas, but I guess until cs3 locks are not implemented for localfs as well this is better than not having locks locally at all...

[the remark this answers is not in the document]
glpatcern: I see, indeed it makes sense for now to keep both implementations until we're sure that all reasonably usable/used storage providers implement the lock API.

I'll have another look at the current code.

glpatcern (changes requested):
I think it's generally all ok, given the need to maintain the previous implementation for a while.
The CodeQL warnings and errors seem to require to be fixed though.

mateusz-garlacz:
I fixed most of the CodeQL warnings. The remaining ones will have to stay, because our tests structure enforce overwriting attributes in subclasses

glpatcern:
I guess it's fine for the overwrite of attributes in subclasses (though it's a bit of an edge case), but the following warnings:
`Missing call to `__init__` during object initialization`
Correspond to missing `super().__init__()` calls. IMHO these must be fixed...

mateusz-garlacz:
all checks are green now

PR #212, opened 2022-12-05
Added duplicates filter to list_shares

mateusz-garlacz:


diocas:
Can you provide some info for this PR?
Also, can you justify why we would need to get the full list of shares (with duplicates)? If there is none, I guess we don't need to overcomplicate it.

mateusz-garlacz:
Sorry for lack of description. This is a cleanup task discovered during Piotr's work. The truth is I can't see a case for getting duplicated shares now, but having hardcoded always enabled filtering seemed not clean. Besides if OCM will be enabled, current solution won't filter OCM shares correctly.

diocas:
I do not understand why REVA doesn't give us already the non duplicated list (ping @glpatcern), but fine.. We can merge.

glpatcern:
> I do not understand why REVA doesn't give us already the non duplicated list (ping @glpatcern), but fine.. We can merge.

I understand that the list includes an entry for each (`path`, `sharee`) tuple, so yes the same path may be shared to multiple people and this appears as multiple shares.

mateusz-garlacz:
Clarification: reva does not return duplicates. I meant that when sharing one file to multiple people, the share list containes all the shares with the same file. For our use (to present in Shared By Me) we need in fact a list of shared files. During mapping we already ensured that one file will appear only once on the list, but the previous solution was not elegant and incorrect, because it compared file paths not resource ids.

diocas:
Ah, yes, ofc.. Sorry.
I was wondering if a share should have a list of sharees (created by the grouping of shares)? I guess we don't need that at the moment in UI nor are planing to. But then again, we also don't need the api to have the `filter_duplicates` param, and it would be unnecessary if we kept the info when grouping.
But for now, let's just merge.

diocas (approved):

PR #221, opened 2023-03-30
Create a copy of a locked file

piotrWichlinskiSoftwaremind:


diocas:
Is this PR ready for review? I still see some things that look like leftovers (files which the only change is an empty line, prints, etc)

cs3api4lab/api/cs3_share_api.py
```diff
@@ -230,7 +230,7 @@ def _shares_received_filter_by_resource(self, path):
         try:
             file_stat = self.file_api.stat_info(path)
         except FileNotFoundError:
-            return sharing.ListReceivedSharesRequest()
+            return sharing.ListSharesRequest()
```
diocas, line 233: This is not really related to this PR, right?

cs3api4lab/api/cs3apismanager.py
```diff
@@ -493,3 +508,37 @@ def list_checkpoints(self, path):
...
+            opaque_id = urllib.parse.unquote(file_info['inode']['opaque_id'])
+            storage_id = urllib.parse.unquote(file_info['inode']['storage_id'])
+            file_info = self.file_api.stat_info(opaque_id, storage_id)
+
+        conflict_file = self.lock_api.resolve_file_path(path)
+        conflict_file_exists = self.file_exists(conflict_file)
+        conflict_file_created = False
+        if not conflict_file_exists:
+            try:
+                file_content = self._read_file(file_info)
```
diocas, line 530: Do we need to read+write? Can't we just copy a file from a location to another?

cs3api4lab/api/cs3apismanager.py
```diff
@@ -493,3 +508,37 @@ def list_checkpoints(self, path):
 
     def delete_checkpoint(self, checkpoint_id, path):
         pass
+
+    @asyncify
+    def create_conflict_file(self, path):
```
diocas, line 513: I'm not a big fan of calling this "conflict file". Those are files that were a cause of a concurrency issue and should not happen. This is more a "copy" or a "clone". And I would have that copy filename have the original filename + the user that created. Nothing more (like timestamps).

By separating these 2 cases, we also make it easier to distinguish if we want to diff in the future (although, we might want to merge in both cases).

cs3api4lab/handlers.py
```diff
@@ -218,7 +231,8 @@ def setup_handlers(web_app, url_path):
         (r"/api/cs3/user", UserInfoHandler),
         (r"/api/cs3/user/claim", UserInfoClaimHandler),
         (r"/api/cs3/user/query", UserQueryHandler),
-        (r"/api/cs3/user/home_dir", HomeDirHandler)
+        (r"/api/cs3/user/home_dir", HomeDirHandler),
+        (r"/api/cs3/locks/create_conflict_file", LockHandler),
```
diocas, line 235: I don't see a good reason to have a dedicated backend api endpoint for this. There's a way to create a copy of a file, so why don't we use it?

From the contents manager docs (for POST):
```
        """Create a new file in the specified path.

        POST creates new files. The server always decides on the name.

        POST /api/contents/path
          New untitled, empty file or directory.
        POST /api/contents/path
          with body {"copy_from" : "/path/to/OtherNotebook.ipynb"}
          New copy of OtherNotebook in path
        """
```

So we can just ask for a copy of original file + current username (which we should have in the frontend as well, right?).

diocas: Looks like we're missing the copy function in the cm... This is important because the UI also provides that option.

diocas: I believe the copy function comes from the parent class that we inherit from jupyter_server... But they implement in a similar way (which is getting and saving the file). We could prob make it more efficient, but let's keep the default behaviour.

masi-io: Just for keeping the records: we agreed that a dedicated backend api endpoint can stay.

Rationale for a separate backend endpoint:
- encapsulation of the logic for copyting the file (and recognizing the copy by filename)
- keeping the logic in backend (so it is easier to update in case some edge cases disturb and need action)
- moving the logic to FE is not really simpler -> getting the username to FE (user in the FE – might not be the same as one in Reva, implication of which need further analysis)
- for the future development: if handling some specific scenarios proves that persistence of copy information is necessary, this can be handled here

src/drive.ts
```diff
@@ -221,15 +242,58 @@ export async function CS3ContainerFiles(
...
+    if (result.button.caption === 'DIFF') {
+      await requestAPI('/api/cs3/locks/create_conflict_file', {
+        method: 'POST',
+        body: JSON.stringify({
+          file_path: path
+        })
+      })
+        .then(async response => {
+          const diffResponse = response as ConflictFileResponse;
+          if (diffResponse.conflict_file_path) {
```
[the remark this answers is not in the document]
piotrWichlinskiSoftwaremind: Other files stay open, this only closes the locked file and opens a copy.

cs3api4lab/api/cs3apismanager.py
```diff
@@ -493,3 +508,37 @@ def list_checkpoints(self, path):
...
+                self.file_api.write_file(conflict_file_path, file_content, self.cs3_config.endpoint, None)
+
+                conflict_file_created = True
+            except Exception:
+                self.log.info('Could not create a conflict file from original %s', path)
+
+        return {
+            'conflict_file_path': conflict_file_path,
+            'conflict_file_created': conflict_file_created,
+            'conflict_file_exists': self.file_exists(conflict_file)
```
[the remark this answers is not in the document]
piotrWichlinskiSoftwaremind: It's about frontend logic and what to do with files, but at the moment we don't need it so I've removed that option

cs3api4lab/api/cs3_share_api.py
piotrWichlinskiSoftwaremind: It's not, and it's actually incorrect, it should be ListReceivedSharesRequest

cs3api4lab/api/cs3apismanager.py
```diff
@@ -495,3 +510,37 @@ def list_checkpoints(self, path):
...
+            try:
+                file_content = self._read_file(file_info)
+                conflict_file_path = FileUtils.normalize_path(notebook_container + '/' + conflict_file)
+                conflict_file_path = FileUtils.check_and_transform_file_path(conflict_file_path)
+
+                self.file_api.write_file(conflict_file_path, file_content, self.cs3_config.endpoint, None)
+
+                conflict_file_created = True
+            except Exception:
+                self.log.info('Could not create a conflict file from original %s', path)
```
diocas, line 541: Not "a conflict"

diocas: Maybe rename the variables as well to avoid misunderstandings.

diocas (changes requested):
Very small change (should take 5min) and I'll merge. So feel free to start rebasing.

PR #223, opened 2023-06-01
checkpoints: listing and reverting a file

piotrWichlinskiSoftwaremind:


cs3api4lab/api/cs3apismanager.py
```diff
@@ -503,10 +503,11 @@ def create_checkpoint(self, path):
         return {'id': 'checkpoint', 'last_modified': "0"}
```
diocas, line 503: Doesn't this create a wrong entry in the UI?

cs3api4lab/api/cs3_file_api.py
```diff
@@ -261,6 +261,64 @@ def create_directory(self, path, endpoint=None):
         self.log.debug(
             'msg="Invoked create container" filepath="%s" elapsedTimems="%.1f"' % (path, (tend - tstart) * 1000))
 
+    def list_checkpoints(self, path):
+        print('list path', path)
```
diocas, line 265: Remove print

cs3api4lab/api/cs3_file_api.py
```diff
@@ -261,6 +261,64 @@ def create_directory(self, path, endpoint=None):
...
+        ref = FileUtils.get_reference(path)
+        request =cs3sp.ListFileVersionsRequest(ref=ref)
```
diocas, line 267: spacing

cs3api4lab/api/cs3_file_api.py
diocas: Also below

cs3api4lab/api/cs3_file_api.py
```diff
@@ -261,6 +261,64 @@ def create_directory(self, path, endpoint=None):
...
+        return checkpoints
+
+    def restore_checkpoint(self, checkpoint_id, path):
+        # because the checkpoint url is parsed by regular expression, so we need to replace the '.' sign from key
+        checkpoint_id = str(checkpoint_id).replace('-', '.')
+        if self.config.dev_env: # we need this until https://github.com/cs3org/reva/issues/3927 is fixed
+            checkpoint_id = 'v' + checkpoint_id
+
+        ref = FileUtils.get_reference(path, self.config.endpoint)
+        request = cs3sp.RestoreFileVersionRequest(ref=ref, key=checkpoint_id)
```
diocas, line 298: Can we move this request bellow (inside the if == metadata)? I think it becomes a bit clearer.

cs3api4lab/api/cs3_file_api.py
```diff
@@ -261,6 +261,64 @@ def create_directory(self, path, endpoint=None):
...
+            opaque_id = urllib.parse.unquote(file_info['inode']['opaque_id'])
+            storage_id = urllib.parse.unquote(file_info['inode']['storage_id'])
+            file_info = self.stat_info(opaque_id, storage_id)
+
+        if self.config.locks_api == "metadata":
+            if self.lock_api.is_file_locked(file_info):
+                self.log.error(f"File {path} is locked and cannot be restored")
+                raise IOError(f"File {path} is locked and cannot be restored")
+        elif self.config.locks_api == "cs3":
+            if self.lock_api.is_file_locked(file_info):
```
diocas, line 312: If file is locked, it means that a lock exists and it's not ours, right?
So, the call bellow will always fail, no? Because we should never be able to restore if we're not locking it ourselves (or no lock exists).

Shouldn't we pass the lock id to the restore, only if there's a lock and it's ours?

diocas: Given the comment above, maybe we can simplify this logic a bit.

diocas (commented):
I see that the checkpoints code is embedded in the middle of the contents manager. Usually, Jupyter allows us to set a `checkpoints_class` in the contents manager (that implements the checkpoints api). Would that still work? If not, can we split this so that we use our checkpoints code as default but we can still allow configuration?
Or, also valid, you argue that we should not allow this to be changed and should just use ours all the time?

I imagine that most actions that would be required (like renaming the checkpoints if a file is renamed) are done automatically by Reva, right?
That is why we only need to list and restore, right?

(ps: we inherit from `ContentsManager` but I'm not sure if we use anything from it.. So having that  `checkpoints_class` doesn't change anything..)

I also added some comments in the code.

cs3api4lab/api/cs3_file_api.py
piotrWichlinskiSoftwaremind: > Shouldn't we pass the lock id to the restore, only if there's a lock and it's ours?

This is precisely what we are doing, only if the file is locked and it's not ours where are passing the lock id to the request, the if statement determines if the file is locked and forms additional info (lock id) for the request object. The actual restore call is not in if-else statement. So if there's no lock we're using basic request object formed above and it's used for  "metadata" locks as well as "cs3" locks (when the file is not locked).

cs3api4lab/api/cs3_file_api.py
piotrWichlinskiSoftwaremind: The request object is used for metadata and for cs3 locks as well (for a file without a lock). I've moved it a little lower, just before if statement so it would be more readable.

cs3api4lab/api/cs3apismanager.py
piotrWichlinskiSoftwaremind: The returned value from this endpoint is only used by front-end tests of the original components. The UI uses another endpoint for listing values and ignores the response body of the create_checkpoint  endpoint.

piotrWichlinskiSoftwaremind:
@diocas  I have added a configuration option so there will be a way to replace the checkpoints if necessary. It works just as authentication.

> I imagine that most actions that would be required (like renaming the checkpoints if a file is renamed) are done automatically by Reva, right?
> That is why we only need to list and restore, right?
Yes, exactly, most actions are done by Reva, we need to list and call restore if necessary.

I have checked the standard jupyter settings to replace checkpoints_class but unfortunately, it does not work.

diocas:
CI is broken due to changes to the REVA examples. Will merge and fix it later.

diocas (approved):

PR #230, opened 2023-06-21
New shared by me / with me panels

### LINKED ISSUE #225: Wrong Shared with me shares and bad selection
In the Shared with me tab panel, I see a ton of shares that I don't see in my "approved" CERNBox list... Meaning that we are showing rejected shares in Shared with me... Why?

Also, I see duplicated shares (they are probably different, but have the same name), which then causes this on selection (I just clicked in one of them):

Screenshot 2023-06-07 at 16 10 49

piotrWichlinskiSoftwaremind:
This PR:
 - adds new shared by me / shared with me panel based on pending share panels
 - fixes share info for received shares
 - resolves #225

### DIFF
--- a/.github/workflows/CI/revad-compose.yml
+++ b/.github/workflows/CI/revad-compose.yml
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   revad:
-    image: cs3org/revad # 'latest' tag is temporarily not available
+    image: cs3org/revad:v1.21.0 # 'latest' tag is temporarily not available
     container_name: revad
     ports:
       - "19000-19001:19000-19001"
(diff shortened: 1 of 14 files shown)


diocas (approved):

diocas:
Will approve, merge and, in case something needs fixing, will do it myself.

.github/workflows/CI/revad-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   revad:
-    image: cs3org/revad # 'latest' tag is temporarily not available
+    image: cs3org/revad:v2.0.0 # 'latest' tag is temporarily not available
```
glpatcern, line 4: Note that reva v2.x tags are used by ownCloud for OCIS, and are significantly different (and incompatible) with v1.x tags, in particular from a StorageProvider API perspective. The current "latest" reva tag is 1.24.

